PR #1, opened 2021-01-08
Development

theresenieva:
Create a basic homepage.

Created the Marker.java only to save the directories.

### DIFF
--- a/prototype/src/main/java/com/google/sps/data/Marker.java
+++ b/prototype/src/main/java/com/google/sps/data/Marker.java
@@ -1,0 +1,1 @@
+// Marker.java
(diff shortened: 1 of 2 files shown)


theresenieva:
@googlebot I signed it!

vivianho18 (commented):
LGTM 👍

prototype/src/main/java/com/google/sps/data/Marker.java
```diff
@@ -0,0 +1 @@
+// Marker.java
```
anna-giang, line 1: (Nitpick) Perhaps this file could have been added when the saving of the directories was actually implemented, because it is essentially an empty file, and seems to be unrelated to other modified files in this commit.

anna-giang: Otherwise LGTM too! 😊

theresenieva: True, I've split the commit into the two parts like you suggested.

PR #3, opened 2021-01-08
Create meeting sequence step 1 - Set meeting name and add guests

anna-giang:
Add a page to allow users to set a meeting name and add meeting guests when they create a meeting.

### DIFF
--- a/prototype/src/main/webapp/create-meeting-step1.html
+++ b/prototype/src/main/webapp/create-meeting-step1.html
@@ -1,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+  <!-- Create Meeting Step 1: Set meeting name and add guests -->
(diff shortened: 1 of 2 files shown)


anna-giang:
I accidentally rebased my branch onto master, which is why there is two copies of each commit and the 'Merge branch...' commit (sorry! 😅)

theresenieva:
LGTM

theresenieva (commented):
LGTM. I used git reset to remove unwanted commits from my branch, maybe that could be useful for you?

vivianho18 (commented):
LGTM - only thing is I don't think you need to bother too much with user input checks at the moment for the prototype.

BipBopBoop (approved):
lgtm

PR #7, opened 2021-01-08
Create meeting sequence Step 2

anna-giang:
Prototype for Step 2 of the Create Meeting sequence.
Users will use this page to select a meeting duration, and to choose whether they want to propose times manually or via Google Calendar.

### DIFF
--- a/prototype/src/main/webapp/create-meeting-step2.html
+++ b/prototype/src/main/webapp/create-meeting-step2.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+  <!-- Create Meeting Step 2: Set meeting duration and select Google Calendar or Manual time proposal -->
+  <head>
+    <meta charset="UTF-8" />
+    <title>Create Meeting</title>
+    <script src="createMeeting.js"></script>
+  </head>
+  <body>
+    <div>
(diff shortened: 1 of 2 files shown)


prototype/src/main/webapp/create-meeting-step2.html
```diff
@@ -0,0 +1,32 @@
...
+  </head>
+  <body>
+    <div>
+      <h1>Find a Meeting Time</h1>
+      <h2>Step 2</h2>
+      <form>
+        <div>
+          <label for="duration-mins">Meeting Duration (minutes)</label>
+          <!-- maximum meeting length = 1440 mins (24 hours/all-day) -->
+          <input type="number" name="duration-mins" min="1" max="1440" />
```
vivianho18, line 17: Could be better if the user could input meeting duration as number of hours and minutes - maybe an alpha feature

vivianho18 (commented):
Looks good :)

theresenieva (commented):
Yup, lgtm

ghost (approved):

prototype/src/main/webapp/create-meeting-step2.html
anna-giang: Yep, noted :)

PR #8, opened 2021-01-10
Create upcoming meetings page

vivianho18:
Each meeting links to view-meeting.html page

### DIFF
--- a/prototype/src/main/webapp/upcoming-meetings.html
+++ b/prototype/src/main/webapp/upcoming-meetings.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>GoMeet Upcoming Meetings</title>
+  <script src="script.js"></script>
+  <link rel="stylesheet" href="style.css">
+  </head>
+  <body>
+    <h1 style="text-align:center;">GoMeet Upcoming Meetings</h1>
+    <ul>
+        <h4 style="text-align:center;"><a href="view-meeting.html">[Meeting Name 1]</a></h3>
+        <h4 style="text-align:center;"><a href="view-meeting.html">[Meeting Name 2]</a></h3>
+        <h4 style="text-align:center;"><a href="view-meeting.html">[Meeting Name 3]</a></h3>
+    </ul>
+  </body>
+</html>
--- a/prototype/src/main/webapp/upcoming-meetings.html
+++ b/prototype/src/main/webapp/upcoming-meetings.html
@@ -8,9 +8,9 @@
   <body>
     <h1 style="text-align:center;">GoMeet Upcoming Meetings</h1>
     <ul>
-        <h4 style="text-align:center;"><a href="view-meeting.html">[Meeting Name 1]</a></h3>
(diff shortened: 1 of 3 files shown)


prototype/src/main/webapp/upcoming-meetings.html
```diff
@@ -0,0 +1,16 @@
...
+<html>
+<head>
+  <title>GoMeet Upcoming Meetings</title>
+  <script src="script.js"></script>
+  <link rel="stylesheet" href="style.css">
+  </head>
+  <body>
+    <h1 style="text-align:center;">GoMeet Upcoming Meetings</h1>
+    <ul>
+        <h4 style="text-align:center;"><a href="view-meeting.html">[Meeting Name 1]</a></h3>
```
anna-giang: Nit: 2 spaces for indentation

anna-giang:
lgtm 😊

theresenieva (commented):
lgtm

prototype/src/main/webapp/upcoming-meetings.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>GoMeet Upcoming Meetings</title>
+  <script src="script.js"></script>
```
ghost: Which script is this referring to? I remember telling Anna that script.js was a bad name, so this might have changed.

prototype/src/main/webapp/upcoming-meetings.html
```diff
@@ -0,0 +1,16 @@
...
+  <link rel="stylesheet" href="style.css">
+  </head>
+  <body>
+    <h1 style="text-align:center;">GoMeet Upcoming Meetings</h1>
+    <ul>
+      <h4 style="text-align:center;"><a href="view-meeting.html">[Meeting Name 1]</a></h3>
+      <h4 style="text-align:center;"><a href="view-meeting.html">[Meeting Name 2]</a></h3>
+      <h4 style="text-align:center;"><a href="view-meeting.html">[Meeting Name 3]</a></h3>
```
ghost: For these `h4`s consider using a class for all of these, then add this text-align style into the css file. This means that if you want to change how the list of meetings looks, you don't have to change all of these elements.

prototype/src/main/webapp/upcoming-meetings.html
vivianho18: I removed this line since the html doesn't actually refer to a javascript file.

prototype/src/main/webapp/upcoming-meetings.html
vivianho18: Done! :)

ghost: This means that every `<h4>` will be centered - if that's what you want that's ok, but if you only want these ones to be centered, you could put `class="foo"` in the h4 tag and then have the css rule apply to `.foo`.

ghost (approved):

prototype/src/main/webapp/upcoming-meetings.html
vivianho18: Changed it to class="center" and replaced the h4 styling with a "center" class

PR #10, opened 2021-01-10
View meeting locations

theresenieva:
Create page for when attendees are viewing an event's proposed locations. 

This page allows viewers to view locations, see number of votes vs. location and vote on locations. I have chosen to have the bargraph and the map displayed at the same time as this would work well with the real-time updating in the final product.


prototype/src/main/webapp/chooseLocation.html
```diff
@@ -9,6 +9,12 @@
   <body onload="createMap();">
     <h1>Meeting Locations: </h1>
     <p>Click on the map to propose a new location.</p>
-    <div id="map"></div>
+    <div id="map"></div></br>
+    <button id="myButton" class="float-left submit-button" >Send Event</button>
+    <script type="text/javascript">
+    document.getElementById("myButton").onclick = function () {
+        location.href = "/vote-locations.html";
```
vivianho18: nit: 2 spaces for indentation

vivianho18 (commented):
LGTM!

anna-giang:
LGTM too! 😊

prototype/src/main/webapp/chooseLocation.html
```diff
@@ -9,6 +9,12 @@
   <body onload="createMap();">
     <h1>Meeting Locations: </h1>
     <p>Click on the map to propose a new location.</p>
-    <div id="map"></div>
+    <div id="map"></div></br>
+    <button id="myButton" class="float-left submit-button" >Send Event</button>
```
ghost: What are these classes for?

prototype/src/main/webapp/chooseLocation.js
```diff
@@ -28,7 +37,8 @@ function createMarkerForDisplay(lat, lng, content) {
     '<div id="bodyContent">' +
     "<p>Number of Votes: 1</br>" +
     "Voters: You</br></p>" +
-    "<button>VOTE</button><br/><br/>"
+    "<input type=\"checkbox\" id=\"vote\" name=\"vote\" value=\"vote\">" +
+    "<label for=\"vote\"> VOTE</label><br></br>" +
```
ghost: I would use single quotes here (and also change the rest of the strings here to use single quotes) so you don't have to keep escaping the double quotes. Alternatively, use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals), and you won't need the concatenations anymore.

ghost (approved):

prototype/src/main/webapp/chooseLocation.html
theresenieva: It was from an example online but I don't think it is necessary, I will remove it

prototype/src/main/webapp/chooseLocation.js
theresenieva: Thanks Austin - I'll update it :)

PR #11, opened 2021-01-11
Create meeting step 3

anna-giang:
Prototype for manual input of proposed times for the meeting - Step 3 of the create meeting sequence.

### DIFF
--- a/prototype/src/main/webapp/createMeeting.js
+++ b/prototype/src/main/webapp/createMeeting.js
@@ -42,14 +42,14 @@ function removeMeetingGuest(guest) {
  * Add a meeting time datetime-local input to the page, enclosed in 
  * a 'li' element.
  * The element appended to the DOM will have an id of 'meeting-time-NUMBER', 
- * where NUMBER equal to value of MEETING_TIMES_ADDED (which is incremented
+ * where NUMBER equal to value of meetingTimesAdded (which is incremented
  * each time this function is called). This is to identify the meeting time input
  * so that users may remove the meeting time. See removeMeetingTime().
  */
 function addMeetingTime() {
   let listItem = document.createElement('li');
-  listItem.id = 'meeting-time-' + MEETING_TIMES_ADDED; 
-  MEETING_TIMES_ADDED += 1;
+  listItem.id = 'meeting-time-' + meetingTimesAdded; 
+  meetingTimesAdded += 1;
 
   let newTimeInput = document.createElement('input');
   newTimeInput.type = 'datetime-local';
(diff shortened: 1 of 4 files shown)


anna-giang:
Rebased the branch onto `main` before submitting pull request, hence the 'duplicated' commits.

anna-giang:
Screen Shot 2021-01-12 at 9 15 56 am

Screenshot of the added HTML page `create-meeting-step-3.html` 😊

theresenieva (commented):
Yup, lgtm

vivianho18 (commented):
lgtm!

prototype/src/main/webapp/createMeeting.js
```diff
@@ -35,3 +37,47 @@ function addMeetingGuest() {
...
+ * Add a meeting time datetime-local input to the page, enclosed in 
+ * a 'li' element.
+ * The element appended to the DOM will have an id of 'meeting-time-NUMBER', 
+ * where NUMBER equal to value of MEETING_TIMES_ADDED (which is incremented
+ * each time this function is called). This is to identify the meeting time input
+ * so that users may remove the meeting time. See removeMeetingTime().
+ */
+function addMeetingTime() {
+  let listItem = document.createElement('li');
+  listItem.id = 'meeting-time-' + MEETING_TIMES_ADDED;
```
ghost: Did you mean `meetingTimesAdded`? Same for all the other occurences of this.

prototype/src/main/webapp/createMeeting.js
```diff
@@ -35,3 +37,47 @@ function addMeetingGuest() {
...
+  // delete button associated with this meeting time
+  let removeTimeButton = document.createElement('button');
+  removeTimeButton.innerHTML = '-';
+  removeTimeButton.addEventListener('click', function() {removeMeetingTime(listItem.id);});
+
+  listItem.appendChild(newTimeInputLabel);
+  listItem.appendChild(newTimeInput);
+  listItem.appendChild(removeTimeButton);
+
+  let timeListDiv = document.getElementById('meeting-time-list')
```
ghost: nit: semicolon

ghost (commented):
Something to think about is that the user may want to delete the first time input when there are more than one. For the prototype this is probably fine, but think about how you could show the remove buttons when there are multiple inputs, and then hide them when there is only one.

prototype/src/main/webapp/createMeeting.js
anna-giang: Oh no, I forgot to change it everywhere 😅 I've changed it now, thanks for picking that up!

ghost (approved):

PR #15, opened 2021-01-14
Create meeting step1

vivianho18:
Front end development for the first step in the meeting creation process

### DIFF
--- a/GoMeet/README.md
+++ b/GoMeet/README.md
@@ -1,0 +1,3 @@
+# GoMeet Web App 
+
+Start Development: 14/01/2021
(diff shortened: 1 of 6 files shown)


suprsrsly (commented):
Let's talk about tests here - I think a couple of tests could be added:
- a basic one that exercises the usual case
- one where guestName is empty

The tests should validate the changes to the DOM that are intended by the functions. WDYT?

Going with that thought, if we don't dependency inject the document object into the functions we'll have to find a way to reset `document` state at the start of each test. Maybe doing a deep copy of it and then overriding the document object before calling the test functions?

PR #16, opened 2021-01-15
Save meeting location

theresenieva:
Add a map that stores location information from user's click and input to the database.

### DIFF
--- a/GoMeet/src/main/java/com/google/sps/data/Location.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Location.java
@@ -1,0 +1,45 @@
+/** Represents a marker on the map. */
+package com.google.sps.data;
+import java.util.Collection;
+
+public class Location {
+
+  private final String title;
+  private final double lat;
+  private final double lng;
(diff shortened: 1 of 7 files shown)


theresenieva:
Not too sure whether `f120cf4`, `d8d3b06`, `b3d00f3`, `e8fcc42` (so all the documentation fixing commits) should be squashed into one.

GoMeet/src/main/java/com/google/sps/data/Location.java
```diff
@@ -0,0 +1,45 @@
...
+  private final double lat;
+  private final double lng;
+  private final String note;
+  private int voteCount;
+  private Collection<String> voters;
+
+  public Location(String title, double lat, double lng, String note) {
+    this.title = title;
+    this.lat = lat;
+	this.lng = lng;
```
vivianho18: nit: fix indentation

vivianho18 (commented):
Should probably include testing for the javacsript file. Other than that lgtm 👍

PR #18, opened 2021-01-18
Create meeting: set duration and meeting time proposal method

anna-giang:
Added functionality for the user to set the meeting duration and time proposal method (manually or via Google Calendar - to be added later).


GoMeet/src/main/webapp/test/spec/specCreateMeetingShared.js
```diff
@@ -0,0 +1,190 @@
+class MockHTMLCollection {
+  /**
+   * Construct a MockHTMLCollection object, a fake that mocks
```
suprsrsly, line 3: nit: fakes and mocks are often different things... https://stackoverflow.com/a/55030455

GoMeet/src/main/webapp/js/createMeetingShared.js
```diff
@@ -0,0 +1,48 @@
...
+        radioChecked.add(input.name);
+        sessionStorage.setItem(key, value);
+      }
+    }
+    else { // all other types of inputs
+      if (value !== '') {
+        sessionStorage.setItem(key, value);
+      }
+      else {
+        alert(FILL_ALL_FIELDS_ALERT);
```
suprsrsly: Hmm, I'm not sure if it's the right thing to add UI calls at this level. It might be better to return a status code and produce the UI at a higher layer instead. In particular, this function might be fiddlier to test if it's producing UI inside your test harness.

suprsrsly:
BTW, it's OK to squash together the commits that have the initial implementation of saveMeeting() and the tests for it. Makes things a little easier to review and associates the two more easily.

GoMeet/src/main/webapp/js/createMeetingStep2.js
```diff
@@ -0,0 +1,41 @@
...
+ * valid.
+ * The duration should at least be 1 minute, and hours is a value 0-23
+ * and minutes 0-59. Alerts the user to input valid value(s)
+ * @returns true if duration is valid, false otherwise.
+ */
+function validateDuration() {
+  let hours = document.getElementById('duration-hours').value;
+  let mins = document.getElementById('duration-mins').value;
+  if (hours === '' || mins === '') {
+    alert(FILL_ALL_FIELDS_ALERT);
```
anna-giang: @suprsrsly Should I also return status codes here as well and alert at a higher layer?

GoMeet/src/main/webapp/js/createMeetingStep2.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+  return true;
+}
+
+/**
+ * Sets the duration and meeting time finding method for this meeting,
+ * by saving the duration and method to session storage, and redirects
+ * to the next step in the create meeting sequence.
+ */
+function saveDurationAndMethod() {
```
anna-giang: @suprsrsly This function executes both saveMeeting and validateDuration and sends the user to the next page, so I'm thinking if I do implement the status codes, they will be handled here and the UI alerts will be created here. When you say this is the 'higher layer' where I should do it at? And would I have to test this function (thus the UI alerts) as well?

GoMeet/src/main/webapp/js/createMeetingShared.js
anna-giang: Is there a best practices on defining my own error codes? I'm thinking they will be integers (stored in a `const` variable), but what number do I start from 😅?

suprsrsly: Maybe an Error object is the right thing? http://go/mdn/JavaScript/Reference/Global_Objects/Error

suprsrsly: Though as status code as a string might be OK too since it's a lot more lightweight.

GoMeet/src/main/webapp/js/createMeetingStep2.js
suprsrsly: yeah, I think so. Displaying alerts inside a validateFoo() function doesn't seem right because it might complicate testing.

GoMeet/src/main/webapp/js/createMeetingShared.js
anna-giang: Yeah, I was actually just looking into [throwing](https://www.w3schools.com/js/js_errors.asp) errors and you messaged - I think I can maybe do something like `throw(error message string)` and then catch and handle it at a higher level

suprsrsly: Hmm.. the spirit of the function is to validate some data, not to try an action that could fail. I think it's OK to do status codes. If you decide to use an exception, I would avoid throwing anything that's not an Error object because the js style guide suggests otherwise: go/js-style#features-exceptions

anna-giang: I decided to go with `Error` objects, because if i returned a status code, I will need some way to map this status code to an error message in order to alert() the user, which would involve a lot of hard-coding that would probably be a pain to modify later on. Returning the entire error message string didn't seem like good practice either.

Currently, I just use `throw(new Error(error message))`, but I could also define my own error classes that are subclasses of `Error`. But the only thing I will modify in those subclasses will be their error message, so I don't know if it is worth it - what do you think?

ghost:
Two advantages to using classes off the top of my head:

1. You can add supplementary fields (e.g. user id) to the error

2. Easier to handle - i.e. checking for a specific error is a instanceof
call rather than a string comparison

anna-giang:
> BTW, it's OK to squash together the commits that have the initial implementation of saveMeeting() and the tests for it. Makes things a little easier to review and associates the two more easily.

I have squashed the commits together as you mentioned @suprsrsly

GoMeet/src/main/webapp/js/createMeetingShared.js
```diff
@@ -0,0 +1,45 @@
...
+      if (!radioGroups.has(input.name)) {
+        radioGroups.add(input.name);
+      }
+      // If the radio button is checked, add to checked set
+      if (input.checked) {
+        radioChecked.add(input.name);
+        sessionStorage.setItem(key, value);
+      }
+    }
+    else { // all other types of inputs
```
vivianho18: nit: according to Google Javascript style guide "there is no line break after the brace if it is followed by else, catch, while, or a comma, semicolon, or right-parenthesis"

GoMeet/src/main/webapp/js/createMeetingShared.js
```diff
@@ -0,0 +1,45 @@
...
+      if (value !== '') {
+        sessionStorage.setItem(key, value);
+      }
+      else {
```
vivianho18: nit: same as above ^

GoMeet/src/main/webapp/js/createMeetingStep2.js
```diff
@@ -0,0 +1,41 @@
...
+  try {
+    if (validateDuration() && saveMeeting()) {
+      location.href = 'create-meeting-step2-manual.html';
+    }
+  }
+  catch(err) {
```
vivianho18: nit: same thing with the style guide

vivianho18 (commented):
nit: for JavaScript files, from the Google JavaScript guide: 'File names must be all lowercase and may include underscores (_) or dashes (-), but no additional punctuation.' I've used "-" in my files because we followed the same format for the HTML files too :) (This was Therese's comment from my PR)

GoMeet/src/main/webapp/js/createMeetingStep2.js
suprsrsly: I think displaying the alerts here is the right thing because it's called as an onclick= method from the webpage.

anna-giang: Okay, that makes sense!

GoMeet/src/main/webapp/js/createMeetingShared.js
anna-giang: Yep, fixed!

GoMeet/src/main/webapp/js/createMeetingStep2.js
```diff
@@ -0,0 +1,41 @@
...
+ * valid.
+ * The duration should at least be 1 minute, and hours is a value 0-23
+ * and minutes 0-59. Alerts the user to input valid value(s)
+ * @returns true if duration is valid, otherwise throws an Error.
+ */
+function validateDuration() {
+  let hours = document.getElementById('duration-hours').value;
+  let mins = document.getElementById('duration-mins').value;
+  if (hours === '' || mins === '') {
+    throw(new Error(BLANK_FIELDS_ALERT));
```
suprsrsly: I recommend discussing with your teammates about establishing an agreed-upon set of rules when functions should and should not throw(). To contrast, internally at Google, we don't use exceptions in e.g., C++ code, for various practical purposes.

suprsrsly (approved):
Great testing!

GoMeet/src/main/webapp/js/createMeetingStep2.js
anna-giang: Okay, will do!

anna-giang:
Just rebased on to main, hence forced update

PR #21, opened 2021-01-21
Step1 (saving to session storage)

vivianho18:
Save details to session storage

### DIFF
--- a/GoMeet/src/main/webapp/js/create-meeting-step1.js
+++ b/GoMeet/src/main/webapp/js/create-meeting-step1.js
@@ -50,8 +50,10 @@ function validateEmails(doc) {
 
   for (let i = 0; i < allInputs.length; i++) {
     if (allInputs.item(i).type === 'email') {
-      if (!isValidEmail(allInputs.item(i).value)) {
-        return false; 
+      if (allInputs.item(i).value === '') {
+        throw(new Error(BLANK_FIELDS_ALERT)); 
+      } else if (!isValidEmail(allInputs.item(i).value)) {
+        throw(new Error(INVALID_EMAILS_ALERT));
       }
     }
   }
@@ -66,4 +68,19 @@ function validateEmails(doc) {
 function isValidEmail(address) {
   const regex = /^(([^<>()[\]\.,;:\s@\"]+(\.[^<>()[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i;
   return regex.test(address.toLowerCase());
+}
+
+/**
+ * Saves the meeting name, host email and guest emails to session
+ * storage, and redirects to the next step in the create meeting sequence.
+ * @param {Document} doc the document object containing the user inputs
+ */
+ function saveStepOneToSessionStorage(doc) {
+  try {
+    if (validateEmails(doc) && saveMeeting()) {
+      location.href = 'create-meeting-step2.html';
+    }
+  } catch(err) {
+    alert(err.message);
+  }
 }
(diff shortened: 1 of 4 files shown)


GoMeet/src/main/webapp/js/create-meeting-step1.js
```diff
@@ -39,3 +39,48 @@ function addGuest(doc) {
...
+
+/**
+ * Checks whether all the email input elements on the document are valid
+ * Note: Server side checks will also be made to validate emails 
+ * @param {Document} doc the document object containing the email inputs
+ */
+function validateEmails(doc) {
+  let allInputs = doc.getElementsByTagName('INPUT');
+
+  for (let i = 0; i < allInputs.length; i++) {
```
ghost: Look into [querySelector](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) - you should be able to match all inputs that have type "email" with something like `doc.querySelector('input[type="email"]')`

GoMeet/src/main/webapp/test/spec/spec-create-meeting-step1.js
```diff
@@ -69,3 +69,22 @@ describe ("Remove guest function", function() {
     expect(numGuests).toBe(0); 
   });
 });
+
+describe ("isValidEmail", function() {
```
theresenieva: nit: Ordinary string literals are delimited with single quotes (').

theresenieva (approved):
one style change, otherwise lgtm

GoMeet/src/main/webapp/js/create-meeting-step1.js
```diff
@@ -39,3 +39,48 @@ function addGuest(doc) {
...
+  }
+  return true; 
+}
+
+/**
+ * Checks whether the given email address is valid according to the 
+ * regular expression pattern 
+ * @param {String} address the email address to be validated 
+ */
+function isValidEmail(address) {
```
ghost: Hmm, ideally, the point of having `input` elements with `type=email` is that the browser should handle this for you. Check out [this](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation)? The main thing to note is that they have their input fields within one `form` element, and also the button within the form element too, which should validate the inputs when pressed.

GoMeet/src/main/webapp/test/spec/spec-create-meeting-step1.js
vivianho18: fixed :)

GoMeet/src/main/webapp/js/create-meeting-step1.js
vivianho18: Done :)

GoMeet/src/main/webapp/js/create-meeting-step1.js
vivianho18: I tried implementing it like this however combining the forms into one form broke my addGuest function. Also, I wanted to follow a similar structure to Anna's implementation of the submit button in create-meeting-step-2.html by calling a function that calls both the saveMeeting function and a validateXXX function.

ghost: I think it'd be better to use something like [input.checkValidity()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#methods) on the input element instead - this avoids the need for the complicated regex, and you can still use the same tests, but you have to pass in an input element to this function rather than a string.

vivianho18: Done - thanks for the resource!

ghost: Don't forget to update the tests

vivianho18: Sorry that's my bad - updated now

ghost (approved):
Looks good :)

PR #23, opened 2021-01-22
Manual Time Input Functionality

anna-giang:
Added HTML page and related JavaScript (including test code) to facilitate the user manually inputting proposed meeting times.


GoMeet/src/main/webapp/js/manual-time-proposal.js
```diff
@@ -0,0 +1,163 @@
+const MAX_TIME_INPUTS = 5; // Allow up to 5 proposed times
+var timeInputs = 1; // Current number of time inputs on page
+const MEETING_TIME_NAME_PREFIX = 'meeting-time-'; // The prefix of the sessionStorage key name for meeting times
+var timeNameSuffix = 1; // The suffix appended to MEETING_TIME_NAME_PREFIX to make the name unique
+const MAX_TIME_INPUT_REACHED_ALERT = 'Maximum number of proposed times: ' + MAX_TIME_INPUTS.toString();
```
vivianho18, line 5: Should this be in the error-messages.js file instead?

GoMeet/src/main/webapp/js/manual-time-proposal.js
```diff
@@ -0,0 +1,163 @@
...
+ * this function will do nothing.
+ * @param {Document} document the document that this function operates on.
+ */
+function toggleDeleteButtons(document) {
+  let deleteButtons = document.getElementsByClassName('delete-time-button');
+  for (let i = 0; i < deleteButtons.length; i++) {
+    if (deleteButtons[i].style.display === 'none') {
+      deleteButtons[i].style.display = 'inline';
+    }
+    else {
```
vivianho18: nit: this should be on the same line as the closing bracket for the 'if' statement

GoMeet/src/main/webapp/js/manual-time-proposal.js
```diff
@@ -0,0 +1,163 @@
...
+ */
+function rectifyInputtedTime(elem) {
+  let timeEntered = new Date(elem.value);
+  let now = new Date(getDatetimeNow());
+  // If the time entered has been entered before or is earlier than now
+  if (enteredTimes.has(elem.value) || !(timeEntered > now)) {
+    elem.value = '';
+    alert(INVALID_TIME_ERROR);
+  }
+  else {
```
vivianho18: nit: same as above

vivianho18 (commented):
Looks great :)

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: This particular constant depends on MAX_TIME_INPUTS, so both will need to be moved to error-messages.js, but MAX_TIME_INPUTS is not an error message, so I think that's not very good practice. This message is only intended to be used on this JS page, so I think it would be okay to keep it here.

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: I think my fingers just love to press enter before an else 😅

GoMeet/src/main/webapp/js/manual-time-proposal.js
vivianho18: Ok makes sense!

suprsrsly:
> Looks great :)

Not enough people give positive feedback during code review, this is a great thing to do :)

GoMeet/src/main/webapp/js/manual-time-proposal.js
suprsrsly: One possibility is to move all your constants/config into a "constants-config.js" or similar that gets inherited wherever it's needed, but I leave that up to you to do as you see fit.

GoMeet/src/main/webapp/js/manual-time-proposal.js
```diff
@@ -0,0 +1,161 @@
...
+ * Hides or shows the buttons with className 'delete-time-button'.
+ * Shows buttons if they're hidden, and vice-versa.
+ * Hides buttons by setting buttonElement.style.display to 'none', 
+ * and shows the buttons by setting buttonElement.style.display to
+ * 'inline'.
+ * If no buttons on the page were found with className 'delete-time-button',
+ * this function will do nothing.
+ * @param {Document} document the document that this function operates on.
+ */
+function toggleDeleteButtons(document) {
```
suprsrsly: Just a suggestion: this might be a little fragile. Something could happen where the visibility state gets modified and this function would then be doing the opposite of what it's supposed to do. If I were doing this I'd call it setDeleteButtonVisibility(bool, doc). But it's OK for now.

GoMeet/src/main/webapp/js/manual-time-proposal.js
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+      deleteButtons[i].style.display = 'none';
+    }
+  }
+}
+
+/**
+ * Returns the ISO datestring of the local current date and time (hours and minutes),
+ * with the time zone data trimmed.
+ * Seconds and milliseconds are set to 0.
```
suprsrsly: Why are you trimming the time zone and minutes/seconds?

GoMeet/src/main/webapp/js/manual-time-proposal.js
```diff
@@ -0,0 +1,161 @@
+const MAX_TIME_INPUTS = 5; // Allow up to 5 proposed times
+var timeInputs = 1; // Current number of time inputs on page
+const MEETING_TIME_NAME_PREFIX = 'meeting-time-'; // The prefix of the sessionStorage key name for meeting times
+var timeNameSuffix = 1; // The suffix appended to MEETING_TIME_NAME_PREFIX to make the name unique
+const MAX_TIME_INPUT_REACHED_ALERT = 'Maximum number of proposed times: ' + MAX_TIME_INPUTS.toString();
+const TIME_INPUT_FIELD_INDEX = 1; // The index of the <input> field with its parent div
+var enteredTimes = new Set();
```
suprsrsly, line 7: optional: instead of putting this at global scope in this module, maybe it's worth putting it in global scope in the page instead and passing it into rectifyCurrentTime and addTimeInput() from the HTML.

GoMeet/src/main/webapp/js/manual-time-proposal.js
```diff
@@ -0,0 +1,161 @@
...
+  let timeEntered = new Date(elem.value);
+  let now = new Date(getDatetimeNow());
+  // If the time entered has been entered before or is earlier than now
+  if (enteredTimes.has(elem.value) || !(timeEntered > now)) {
+    elem.value = '';
+    alert(INVALID_TIME_ERROR);
+  } else {
+    enteredTimes.add(elem.value);
+  }
+}
```
suprsrsly, line 180: Optional: stylistically I think this function could be structured like saveTimes(). particularly:

err = addFutureTime(elem.value); // just an example for a name
if (err) {
    alert(err.message)
}

suprsrsly: For testing, it'll decouple needing to test that err.message gets displayed from validating the behavior of addFutureTime().

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: When you say global scope of the page, do you mean between <script> tags in the HTML page?

GoMeet/src/main/webapp/test/spec/spec-manual-time-proposal.js
```diff
@@ -0,0 +1,325 @@
...
+    var inputField = addedInputDiv.children.item(1);
+    expect(inputField.tagName).toEqual('INPUT');
+    expect(inputField.type).toEqual('datetime-local');
+    expect(inputField.name).toEqual(inputName);
+    expect(inputField.min).toEqual(ISO_DATESTRING);
+    expect(inputField.onchange.toString()).toBe('function() {rectifyInputtedTime(this);}');
+
+    var deleteButton = addedInputDiv.children.item(2);
+    expect(deleteButton.tagName).toEqual('BUTTON');
+    expect(deleteButton.innerText).toEqual('-');
```
suprsrsly: optional suggestion: since you care about functionality more than interface, you probably don't need to verify the UI text labels.

suprsrsly (approved):

GoMeet/src/main/webapp/js/manual-time-proposal.js
```diff
@@ -0,0 +1,161 @@
...
+  deleteButton.className = 'delete-time-button';
+  deleteButton.onclick = function() {deleteTimeInput(document, inputField.name);}; 
+  
+  let enclosingDiv = document.createElement('div');
+  enclosingDiv.id = inputField.name; // ID of div = name of input field, to faciliate deleting
+
+  enclosingDiv.appendChild(label);
+  enclosingDiv.appendChild(inputField);
+  enclosingDiv.appendChild(deleteButton);
+  inputDiv.appendChild(enclosingDiv);
```
suprsrsly, line 58: Optional suggestion: Maybe you can use a function to create enclosingDiv.

anna-giang:
Rebased onto main to sync with the newly merged changes.

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: I think we will do some refactoring when everything is merged together to see what's the best way to handle the constants.

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: The time zone is _always_ UTC time. So if I create a Date using `Date(datetimeStr with a time in AEDT)`, the ISO string of that date returned by Date.toISOString() will add a 'Z' to the end even though I meant to have the time in AEDT. Also, the 'min' property of the datetime input can only take in strings without the timezone. 

In terms of the seconds and milliseconds, if they are not set to 0, the datetime input field would include the seconds and milliseconds in the display, which looked a little bit messy. I also thought it would be confusing for user since I don't intend for them to be able set a meeting time to that level of granularity.

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: I will add this to one of the things I will come to refactor 👍

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: I've updated the comments of the function to document this 😊

GoMeet/src/main/webapp/test/spec/spec-manual-time-proposal.js
anna-giang: I've removed the checks, because this would also make it easier for us to change the UI for stylistic purposes later. Thanks for the suggestion!

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: I will add that to my refactoring list 👍

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: Yep, I've done that! It does make the tests more focused on testing _either_ the logic of verifying a future time _or_ the UI stuff 😊

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: Actually, it turned out be a very quick fix, so I've refactored it 😅

GoMeet/src/main/webapp/js/manual-time-proposal.js
anna-giang: Again, this was really quick, and does remove repeated code, so I've refactored it 😊

PR #24, opened 2021-01-22
Store location

theresenieva:
Add a map to the meeting event page.

Interacting with this page allows users to add locations.

### DIFF
--- a/GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
+++ b/GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
@@ -1,0 +1,46 @@
+package main.java.com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import main.java.com.google.sps.data.Location;
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.jsoup.Jsoup;
+import org.jsoup.safety.Whitelist;
+
+/** Handles fetching and saving location data. */
+@WebServlet("/location-data")
+public class LocationServlet extends HttpServlet {
+  /** Accepts a POST request containing a new location. */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) {
+    double lat = Double.parseDouble(request.getParameter("lat"));
+    double lng = Double.parseDouble(request.getParameter("lng"));
+    String note = Jsoup.clean(request.getParameter("note"), Whitelist.none());
+    String title = Jsoup.clean(request.getParameter("title"), Whitelist.none());
+
+    Location location = new Location(title, lat, lng, note);
+    storeLocation(location);
+  }
+
+  /** Stores a location in Datastore. */
+  public void storeLocation(Location location) {
+    Entity locationEntity = new Entity("Location");
+    locationEntity.setProperty("title", location.getTitle());
+    locationEntity.setProperty("lat", location.getLat());
+    locationEntity.setProperty("lng", location.getLng());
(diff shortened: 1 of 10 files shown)


vivianho18 (commented):
LGTM :)

GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -0,0 +1,46 @@
...
+import org.jsoup.Jsoup;
+import org.jsoup.safety.Whitelist;
+
+/** Handles fetching and saving location data. */
+@WebServlet("/location-data")
+public class LocationServlet extends HttpServlet {
+  /** Accepts a POST request containing a new location. */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) {
+    double lat = Double.parseDouble(request.getParameter("lat"));
```
anna-giang, line 26: Will this crash if there was no 'lat' and 'lng' parameter in the request? 
Or if the values in the 'lat' and 'lng' params were not numbers?

GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -0,0 +1,46 @@
...
+    double lng = Double.parseDouble(request.getParameter("lng"));
+    String note = Jsoup.clean(request.getParameter("note"), Whitelist.none());
```
anna-giang, line 28: This `jsoup` is a good idea! 😊

GoMeet/src/main/webapp/js/interactive-map.js
```diff
@@ -0,0 +1,77 @@
...
+  map.addListener('click', (event) => {
+    createLocationForEdit(map, event.latLng.lat(), event.latLng.lng());
+  });
+  return map; 
+}
+
+/** Creates a marker that shows a textbox the user can edit. */
+function createLocationForEdit(map, lat, lng) {
+ 
+  editLocation =
```
anna-giang: Same thought as above: if the map/lat/lng arguments were invalid types/values, what would happen? 
I guess this is not as important as functions in an API for example, since the code will only be called within our own application (and we are expecting valid args), but something to think about if we have time 😊

theresenieva: Yup, I think that's a valid point. I was going to modify my fetchWrapper (Will have to rename the class haha) to be similar to your MeetingTimeDAO (which makes checks before sending the request) and/or doing the checks server-side and returning a status based on whether the request was valid or not.

GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  /** Stores a location in Datastore. */
+  public void storeLocation(Location location) {
+    Entity locationEntity = new Entity("Location");
+    locationEntity.setProperty("title", location.getTitle());
+    locationEntity.setProperty("lat", location.getLat());
+    locationEntity.setProperty("lng", location.getLng());
+    locationEntity.setProperty("note", location.getNote());
+
```
vivianho18, line 42: Do you also need to store the voteCount and voters to datastore?

theresenieva: Yup, I was planning to add this in when I implement the 'voting' feature.

ghost: Optional, but one thing that some people do is they add a `//TODO: ` comment to remind themselves or a reader of future tasks.

GoMeet/src/main/webapp/js/fetch-wrapper.js
```diff
@@ -0,0 +1,10 @@
+/** Wraps the fetch function. */
+class FetchWrapper {
+
+  constructor() {};
+
+  async doPost(servlet, params) {
+    let response = await fetch(servlet, {method: 'POST', body: params});
+    return response;
```
ghost: Just `return await fetch(...)`?

GoMeet/src/main/webapp/js/interactive-map.js
ghost: If you don't want to add in checks now, `TODO` comment also works.

ghost: Add `const`

GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
theresenieva: (Answered in another comment)

PR #25, opened 2021-01-26
MeetingTime JavaScript Data Access Object

anna-giang:
MeetingTimeDAO JavaScript class containing functions to fetch data from, and store data to the Java Servlet with the `/meeting-time` endpoint.

### DIFF
--- a/GoMeet/src/main/webapp/js/meeting-time-dao.js
+++ b/GoMeet/src/main/webapp/js/meeting-time-dao.js
@@ -19,6 +19,7 @@ class MeetingTimeDAO {
     }
     return MeetingTimeDAO.endpoint + queryString.toString();
   }
+  
   /**
    * Fetches the meeting time data associated with the provided
    * MeetingTimeID.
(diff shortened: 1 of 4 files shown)


GoMeet/src/main/webapp/js/meeting-time-dao.js
```diff
@@ -0,0 +1,88 @@
...
+      throw new Error(INSUFFICIENT_REQUEST_PARAM);
+    }
+     
+    if (typeof meetingTimeId !== 'string') {
+      throw new Error(INVALID_PARAM_TYPE);
+    }
+
+    let urlString = MeetingTimeDAO.url({'meetingTimeId': meetingTimeId});
+    let results = await fetch(urlString);
+    let resultsJson = await results.json();
```
vivianho18: could combine this into one line with .then and arrow functions

vivianho18 (commented):
LGTM good testing!

GoMeet/src/main/webapp/js/meeting-time-dao.js
anna-giang: I thought I needed to separate them to mock `fetch` by itself, but turns out I didn't need to 😊 (the tests pass after the change!)

GoMeet/src/main/webapp/js/meeting-time-dao.js
```diff
@@ -0,0 +1,86 @@
...
+
+    if (typeof datetimeStr !== 'string') {
+      throw new Error(INVALID_PARAM_TYPE);
+    }
+
+    // check that the datetime string is in the format: YYY-MM-DDTHH:MM
+    let format = new RegExp(/\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d/);
+    if (!format.test(datetimeStr)) {
+      throw new Error(INVALID_PARAM_VALUE);
+    }
```
thanhdng, line 69: Line 73-76 already check the format of the string, any reason to keep this? It also seems that the regex will accept "2020-15-39T28:00" as a match, so maybe consider removing this check?

anna-giang:
Rebased onto main to sync with newly merged PRs.

theresenieva (approved):
lgtm :)

GoMeet/src/main/webapp/js/meeting-time-dao.js
anna-giang: The Date constructor can take a number of different formats (e.g. "20 Jan 2021" can also initialise a date), but the _only_ type of format I want here is the '2021-01-20T11:40'. 

Yep, it is because the regex will accept invalid dates like the one you mentioned, after I check the format, I need pass the string _again_ to the Date constructor to check that it is a valid date _as well as_ being of the correct format.

Hopefully that makes sense!

thanhdng: The regex here is quite scary, and doesn't actually do its job well (couldn't check the content of the string by itself and still need to depends on Date() constructor, and in fact, it'll match something like `blah blah blah 2020-15-39T28:00 blah blah blah`). 

Can you use something like `^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}$` which is solely used to match the format of the string? Also could you update the comment as well (you missed a Y)?

anna-giang: Yep, I agree - the additional checks are really not necessary if I'm going to pass it to the Date constructor anyway. I have changed it now 👍

thanhdng (approved):
LGTM

anna-giang:
Rebasing on to main before merge.

PR #26, opened 2021-01-27
Login dev

vivianho18:
Requires users to log in with their google account before they can see the meeting event details

### DIFF
--- a/GoMeet/src/main/webapp/js/login-status.js
+++ b/GoMeet/src/main/webapp/js/login-status.js
@@ -1,6 +1,9 @@
 class LoginStatus {
-  static async doGet(servlet) {
-    let promise = await fetch(servlet).then(response => response.json());
+  
+  static endpoint = '/user-status'; 
+
+  static async doGet() {
+    let promise = await fetch(LoginStatus.endpoint).then(response => response.json());
     return promise; 
   }
 }
--- a/GoMeet/src/main/webapp/js/login-status.js
+++ b/GoMeet/src/main/webapp/js/login-status.js
@@ -1,0 +1,6 @@
+class LoginStatus {
+  static async doGet(servlet) {
+    let promise = await fetch(servlet).then(response => response.json());
+    return promise; 
+  }
+}
--- a/GoMeet/src/main/webapp/js/meeting-display.js
+++ b/GoMeet/src/main/webapp/js/meeting-display.js
@@ -4,22 +4,23 @@ function toggleMeetingDisplay() {
     let meetingTimes = document.getElementById('vote-meeting-times'); 
     let meetingLocations = document.getElementById('vote-meeting-locations'); 
     let votesTable = document.getElementById('votes-table'); 
+    let prompt; 
     
     if (loginStatus.loggedIn === 'true') {
       meetingName.style.display = 'block';
       meetingTimes.style.display = 'block';
       meetingLocations.style.display = 'block';
       votesTable.style.display = 'block';
-      let logoutPrompt = '<a href=' + loginStatus.logoutUrl + '>Logout</a>';
-      document.getElementById('login-or-logout-prompt').innerHTML = logoutPrompt;
+      prompt = '<a href=' + loginStatus.logoutUrl + '>Logout</a>';
+      document.getElementById('login-or-logout-prompt').innerHTML = prompt;
     } else {
       meetingName.style.display = 'none';
       meetingTimes.style.display = 'none';
       meetingLocations.style.display = 'none';
       votesTable.style.display = 'none';
-      let loginPrompt = '<p>Please log in to view the meeting event details</p>'
+      prompt = '<p>Please log in to view the meeting event details</p>'
       + '<a href=' + loginStatus.loginUrl + '>Login</a>';
-      document.getElementById('login-or-logout-prompt').innerHTML = loginPrompt;
+      document.getElementById('login-or-logout-prompt').innerHTML = prompt;
     }
   });
 }
--- a/GoMeet/src/main/webapp/js/meeting-display.js
+++ b/GoMeet/src/main/webapp/js/meeting-display.js
@@ -1,0 +1,24 @@
+function toggleMeetingDisplay() {
+  LoginStatus.doGet('/user-status').then((loginStatus) => {
+    let meetingName = document.getElementById('meeting-title');
+    let meetingTimes = document.getElementById('vote-meeting-times'); 
+    let meetingLocations = document.getElementById('vote-meeting-locations'); 
+    let votesTable = document.getElementById('votes-table'); 
+    
+    if (loginStatus.loggedIn === 'true') {
+      meetingName.style.display = 'block';
+      meetingTimes.style.display = 'block';
+      meetingLocations.style.display = 'block';
+      votesTable.style.display = 'block';
+      let logoutPrompt = '<a href=' + loginStatus.logoutUrl + '>Logout</a>';
+      document.getElementById('login-or-logout-prompt').innerHTML = logoutPrompt;
+    } else {
+      meetingName.style.display = 'none';
+      meetingTimes.style.display = 'none';
+      meetingLocations.style.display = 'none';
+      votesTable.style.display = 'none';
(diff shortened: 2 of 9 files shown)


GoMeet/src/main/webapp/test/spec/spec-meeting-display.js
```diff
@@ -0,0 +1,56 @@
...
+    await toggleMeetingDisplay();
+
+    expect(meetingName.style.display).toBe('block');
+    expect(meetingTimes.style.display).toBe('block'); 
+    expect(meetingLocations.style.display).toBe('block'); 
+    expect(votesTable.style.display).toBe('block');
+  });
+
+  it ('Should hide all the meeting details if the user is not logged in', async function() {
+    spyOn(LoginStatus, 'doGet').and.returnValue(new Promise((resolve, rejeect) => {
```
theresenieva: nit: I think you meant reject haha

GoMeet/src/main/webapp/meeting-event.html
```diff
@@ -3,12 +3,21 @@
...
+  <body onload="toggleMeetingDisplay()">
     <h1 class="center">GoMeet</h1>
-    <h2 id="meeting-name"></h2>
-    <div id="vote-meeting-times"></div>
-    <div id="vote-meeting-locations"></div>
-    <div id="votes-table"></div>
+    <p id='login-or-logout-prompt'></p>
+    <h2 id="meeting-title">Meeting Name:</h2>
+    <div id="vote-meeting-times">
+        <h3>Meeting times here</h3>
```
theresenieva: nit: looks like indents are +4 here?

GoMeet/src/main/webapp/js/login-status.js
```diff
@@ -0,0 +1,6 @@
+class LoginStatus {
+  static async doGet(servlet) {
```
theresenieva: Optional: Is this just going to be used to make requests to '/user-status'? In that case, could consider making '/user-status' a static variable in this class (similar to the Meeting-time-DAO). I've done something similar for meeting-location-dao. I think this would make the project more uniform overall.

GoMeet/src/main/webapp/js/meeting-display.js
```diff
@@ -0,0 +1,24 @@
...
+      meetingLocations.style.display = 'block';
+      votesTable.style.display = 'block';
+      let logoutPrompt = '<a href=' + loginStatus.logoutUrl + '>Logout</a>';
+      document.getElementById('login-or-logout-prompt').innerHTML = logoutPrompt;
+    } else {
+      meetingName.style.display = 'none';
+      meetingTimes.style.display = 'none';
+      meetingLocations.style.display = 'none';
+      votesTable.style.display = 'none';
+      let loginPrompt = '<p>You must be logged in to view the meeting event details</p>' + '<a href=' + loginStatus.loginUrl + '>Login</a>';
```
theresenieva: nit: JavaScript code has a column limit of 80 characters.

theresenieva (approved):
Overall, looks good :))

GoMeet/src/main/webapp/js/login-status.js
theresenieva: Also, in future, I think this class would be a good place to add any checks i.e. unexpected responses from the server.

GoMeet/src/main/webapp/test/spec/spec-meeting-display.js
vivianho18: Good spotting thank you!

GoMeet/src/main/webapp/js/login-status.js
vivianho18: Noted :)

anna-giang (approved):
Looks good to me! 😊

GoMeet/src/main/webapp/js/meeting-display.js
```diff
@@ -0,0 +1,25 @@
...
+      document.getElementById('login-or-logout-prompt').innerHTML = logoutPrompt;
+    } else {
+      meetingName.style.display = 'none';
+      meetingTimes.style.display = 'none';
+      meetingLocations.style.display = 'none';
+      votesTable.style.display = 'none';
+      let loginPrompt = '<p>Please log in to view the meeting event details</p>'
+      + '<a href=' + loginStatus.loginUrl + '>Login</a>';
+      document.getElementById('login-or-logout-prompt').innerHTML = loginPrompt;
+    }
```
thanhdng, line 12: Nits: I think we can make this a bit cleaner by doing something like this
`let style = '';  
let prompt = '';
if (loginStatus.loggedIn === 'true') {
  style = 'block';
  prompt = ......  
} else { 
  style = 'none';
  prompt = ....;  
}. 
`

thanhdng (approved):
LGTM with one small nit

GoMeet/src/main/webapp/js/meeting-display.js
vivianho18: Done for the 'prompt' variable. However, if I try to do this to the style lines it seems to just save the value of XXX.style.display in a variable. So when I try to update the value to 'block', it updates the variable value not the actual XXX.style.display value.

thanhdng: Can you try `xxx.style.display = style.valueOf()`?

vivianho18: Screen Shot 2021-02-01 at 12 00 35 pm
is this what you mean? (lines 8 and 12) if so, it doesn't work either :(

thanhdng: I meant something like this
```js
function toggleMeetingDisplay() {
  LoginStatus.doGet().then((loginStatus) => {
    let style = '';
    let prompt = '';

    if (loginStatus.loggedIn === 'true') {
      style = 'block';
      prompt = '<a href=' + loginStatus.logoutUrl + '>Logout</a>';
    } else {
      style = 'none';
      prompt = '<p>Please log in to view the meeting event details</p>'
      + '<a href=' + loginStatus.loginUrl + '>Login</a>';
    }
    
    document.getElementById('meeting-title').style.display = style.valueOf();
    document.getElementById('vote-meeting-times').style.display = style.valueOf();
    document.getElementById('vote-meeting-locations').style.display = style.valueOf();
    document.getElementById('votes-table').style.display = style.valueOf();
    document.getElementById('login-or-logout-prompt').innerHTML = prompt.valueOf();
    
   ```

vivianho18: Sorry I'm not too sure what you mean by this. Is it: 
let styleVar = 'block'; 
if (loggedIn === true) {
  xxx.style.display = styleVar.valueOf();

vivianho18: done thank you :)

thanhdng: LGTM, but why can't we move line 8 and 13 outside of if-else clause?

vivianho18: Sorry my bad - moved!

vivianho18: Do I need to use .valueOf() ? I think it should still work without

thanhdng (approved):
Nice work! :)

GoMeet/src/main/webapp/js/meeting-display.js
thanhdng: I don't think .valueOf() is needed here. The reason why we need .valueOf() of style.display is because it requires to have a value assigned to it (read more here: https://www.w3schools.com/jsref/prop_style_display.asp).

thanhdng: @austinct do you know the reason here?

thanhdng: @austinct do you know the reason here?

ghost: Err I'm not too sure where `valueOf()` comes from in this context - `style`/`prompt` are already strings so calling `valueOf()` on it returns itself. Seems like it's not necessary?

vivianho18: Ok I've removed it and it still passes the tests :)

PR #27, opened 2021-01-27
Display locations

theresenieva:
Displays the location data on the database on the map on the webpage. 

Note: The commits from Jan 22 2021 are part of a PR that hasn't been merged yet.

### DIFF
--- a/GoMeet/src/main/webapp/js/fetch-wrapper.js
+++ b/GoMeet/src/main/webapp/js/fetch-wrapper.js
@@ -7,4 +7,9 @@ class FetchWrapper {
     let response = await fetch(servlet, {method: 'POST', body: params});
     return response;
   }
+
+  async doGet(servlet) {
+    let response = await fetch(servlet);
+    return response;
+  }
 }
--- a/GoMeet/src/main/webapp/js/interactive-map.js
+++ b/GoMeet/src/main/webapp/js/interactive-map.js
@@ -7,6 +7,9 @@ function createMap() {
   map.addListener('click', (event) => {
     createLocationForEdit(map, event.latLng.lat(), event.latLng.lng());
   });
+
+  const fetchWrapper = new FetchWrapper();
+  fetchLocations(map, fetchWrapper);
   return map; 
 }
 
@@ -25,6 +28,18 @@ function createLocationForEdit(map, lat, lng) {
   });
   infoWindow.open(map, editLocation);
   return editLocation;
+}
+
+/** Creates a marker that shows the location's information. */
+async function createLocationForDisplay(map, lat, lng, title) {
+  let displayLocation =
+      new google.maps.Marker({position: {lat: lat, lng: lng}, map: map});
+
+  const infoWindow = new google.maps.InfoWindow({content: title});
+  displayLocation.addListener('click', () => {
+    infoWindow.open(map, displayLocation);
+  });
+  return displayLocation;
 }
 
 /**
@@ -74,4 +89,13 @@ function validateTitle(title) {
   if (title === '') {
     throw (new Error(BLANK_FIELDS_ALERT));
   }
+}
+
+/** Fetches the location data. */
+async function fetchLocations(map, fetchWrapper) {
+  let response = await fetchWrapper.doGet('location-data');
+  let json = await response.json();
+  json.forEach(async (location) => {
+    await createLocationForDisplay(map, location.lat, location.lng, location.title);
+  });
 }
--- a/GoMeet/src/main/webapp/js/interactive-map.js
+++ b/GoMeet/src/main/webapp/js/interactive-map.js
@@ -1,3 +1,13 @@
+/** Initialises the map. */
+function initMap() {
+  // Create the map object
+  let map = createMap();
+  
+  // Add the data from the database to the map
+  const fetchWrapper = new FetchWrapper();
+  fetchLocations(map, fetchWrapper);
+}
+
 /** Creates a map that allows users to add markers. */
 function createMap() {
   let map = new google.maps.Map(
@@ -7,9 +17,6 @@ function createMap() {
   map.addListener('click', (event) => {
     createLocationForEdit(map, event.latLng.lat(), event.latLng.lng());
   });
-
-  const fetchWrapper = new FetchWrapper();
-  fetchLocations(map, fetchWrapper);
   return map; 
 }
 
--- a/GoMeet/src/main/webapp/js/interactive-map.js
+++ b/GoMeet/src/main/webapp/js/interactive-map.js
@@ -39,7 +39,7 @@ function createLocationForEdit(map, lat, lng) {
 }
 
 /** Creates a marker that shows the location's information. */
-async function createLocationForDisplay(map, lat, lng, title) {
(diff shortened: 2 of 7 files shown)


theresenieva:
Change of base branch was because of this: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch#:~:text=Use%20the%20Edit%20button%20when,from%20the%20most%20recent%20commits.

GoMeet/src/main/webapp/js/fetch-wrapper.js
```diff
@@ -7,4 +7,9 @@ class FetchWrapper {
     let response = await fetch(servlet, {method: 'POST', body: params});
     return response;
   }
+
+  async doGet(servlet) {
```
anna-giang, line 10: It might be a good idea to add some comments to specify what the param `servlet` is (i.e. the URL that the target servlet is mapped to).

GoMeet/src/main/webapp/js/interactive-map.js
```diff
@@ -75,3 +99,13 @@ function validateTitle(title) {
     throw (new Error(BLANK_FIELDS_ALERT));
   }
 }
+
+/** Fetches the location data. */
+async function fetchLocations(map, fetchWrapper) {
+  let response = await fetchWrapper.doGet('location-data');
+  let json = await response.json();
+  json.forEach(async (location) => {
```
vivianho18: Could combine lines 105-107 into one line using the arrow functions

vivianho18 (commented):
lgtm! :)

theresenieva:
Note: fetch-wrapper.js has been replaced by MeetingLocationDAO in #29

GoMeet/src/main/webapp/js/fetch-wrapper.js
theresenieva: FetchWrapper.js has been removed from PR #29 onwards but, in future, I will add JSDoc for any parameters that might not be clear :)

GoMeet/src/main/webapp/test/spec/spec-interactive-map.js
```diff
@@ -3,7 +3,7 @@ describe('Create map', function() {
   it ('Should create and return a map object', function() {
     const mapConstructorSpy = spyOn(google.maps, 'Map');
     const mockedMap = jasmine.createSpyObj('Map', ['addListener']);
-    mapConstructorSpy.and.returnValue(mockedMap);
+    mapConstructorSpy.and.returnValue(mockedMap);
```
ghost: Accidental whitespace?

GoMeet/src/main/webapp/js/interactive-map.js
```diff
@@ -75,3 +99,13 @@ function validateTitle(title) {
     throw (new Error(BLANK_FIELDS_ALERT));
   }
 }
+
+/** Fetches the location data. */
+async function fetchLocations(map, fetchWrapper) {
+  let json = await fetchWrapper.doGet('location-data').then(
+      response => response.json());
+  json.forEach(async (location) => {
```
ghost: This doesn't need to by async, and you don't need the await in the next line because `createLocationForDisplay` isn't async.

Although, if `createLocationForDisplay` was async, this would be wrong in a different way - optional reading below that I wrote before realising `createLocationForDisplay` didn't need to be async :upside_down_face: 

> This is subtly wrong - `forEach` calls this arrow function for each value in `json`, but since the arrow function is asynchronous, it won't wait for each to finish, so these arrow functions are being called in the background.
> 
> You want to do something like
> `await Promise.all(json.map((location) => createLocationForDisplay(...)))`
> 
> Breaking this down:
> `map` is like `forEach`, but returns a list of the function's return values (e.g. `[1, 2, 3].map(x => x+1)` returns `[2, 3, 4]`)
> So for each `location` we call `createLocationForDisplay` on it, which, as it is an async function, returns a Promise, so `map()` returns a list of promises. Then `Promise.all` creates a new promise that waits on all of the promises in a list, so that's how you wait for all of the code to execute.

GoMeet/src/main/webapp/js/interactive-map.js
```diff
@@ -27,6 +38,18 @@ function createLocationForEdit(map, lat, lng) {
   return editLocation;
 }
 
+/** Creates a marker that shows the location's information. */
+async function createLocationForDisplay(map, lat, lng, title) {
```
ghost: This doesn't call any asynchronous functions, so doesn't need to be async.

GoMeet/src/main/webapp/js/interactive-map.js
```diff
@@ -75,3 +99,13 @@ function validateTitle(title) {
     throw (new Error(BLANK_FIELDS_ALERT));
   }
 }
+
+/** Fetches the location data. */
+async function fetchLocations(map, fetchWrapper) {
+  let json = await fetchWrapper.doGet('location-data').then(
+      response => response.json());
```
ghost: IMO I wouldn't combine these with `then`. Stick to just using `await` - I think it's easier to read as you have two simple statements rather than one bigger statement.

GoMeet/src/test/java/com/google/sps/LocationServletTest.java
```diff
@@ -90,4 +101,30 @@ public void doPostTest1() {
...
+    Gson gson = new Gson();
+    String expectedJson = gson.toJson(new ArrayList<>(Arrays.asList(LOCATION_A)));
+
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(response.getWriter()).thenReturn(writer);
+
+    new LocationServlet().doGet(request, response);
+    assertTrue(stringWriter.toString().contains(expectedJson));
+  }
```
ghost, line 149: I think this test relies too much on the implementation of `doGet()`, specifically that it uses Gson. JSON can be formatted in multiple ways with whitespace in between - e.g. `["Hello", "World"]` is the same as
```
[
    "Hello",
    "World"
]
```

And this test relies that the function's output is the same as Gson's JSON string output. Suppose you changed the JSON library in the function, and the new library adds whitespace in different locations - the code would still be correct, but the test would not pass.

What I think you want to test here is that doGet() outputs valid JSON and that the JSON contains an object you specify - so you should instead parse the output string with Gson, and then check that the object contains the expected properties.

GoMeet/src/main/webapp/js/interactive-map.js
theresenieva: Just double checking, based on the other comment, this function doesn't need async/await?

theresenieva: > This doesn't need to by async, and you don't need the await in the next line because `createLocationForDisplay` isn't async.

GoMeet/src/main/webapp/js/interactive-map.js
theresenieva: 😂 🙃 it really do be like that sometimes haha. But that was useful anyways - that was what I was trying to do :))).

ghost (approved):
LGTM!

PR #28, opened 2021-01-29
Location vote

theresenieva:
Add functionality to the vote button (Added in PR #27). Clicking on that button will update the location entity in the database.

### DIFF
--- a/GoMeet/src/main/java/com/google/sps/data/Location.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Location.java
@@ -8,15 +8,15 @@ public class Location {
   private final double lat;
   private final double lng;
   private final String note;
-  private int voteCount;
+  private long voteCount;
   private Collection<String> voters;
 
-  public Location(String title, double lat, double lng, String note) {
+  public Location(String title, double lat, double lng, String note, long voteCount) {
     this.title = title;
     this.lat = lat;
     this.lng = lng;
     this.note = note;
-    this.voteCount = 1;
+    this.voteCount = voteCount;
     voters = new HashSet<>();
   }
 
@@ -40,7 +40,7 @@ public class Location {
     return note;
   }
 
-  public int getVoteCount() {
+  public long getVoteCount() {
     return voteCount;
   }
 
--- a/GoMeet/src/main/java/com/google/sps/data/Location.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Location.java
@@ -10,6 +10,7 @@ public class Location {
   private final String note;
   private long voteCount;
   private Collection<String> voters;
+  private String keyString;
 
   public Location(String title, double lat, double lng, String note, long voteCount) {
     this.title = title;
@@ -18,6 +19,17 @@ public class Location {
     this.note = note;
     this.voteCount = voteCount;
     voters = new HashSet<>();
+    this.keyString = null;
+  }
+
+  public Location(String title, double lat, double lng, String note, long voteCount, String keyString) {
+    this.title = title;
+    this.lat = lat;
+    this.lng = lng;
+    this.note = note;
+    this.voteCount = voteCount;
+    voters = new HashSet<>();
+    this.keyString = keyString;
   }
 
   public void addVoter(String voter) {
@@ -46,5 +58,9 @@ public class Location {
 
   public Collection<String> getVoters() {
     return voters;
+  }
+
+  public String getKeyString() {
+    return keyString;
   }
 }
--- a/GoMeet/src/main/java/com/google/sps/data/Location.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Location.java
@@ -8,11 +8,11 @@ public class Location {
   private final double lat;
   private final double lng;
   private final String note;
-  private long voteCount;
+  private int voteCount;
   private Collection<String> voters;
   private String keyString;
 
-  public Location(String title, double lat, double lng, String note, long voteCount) {
+  public Location(String title, double lat, double lng, String note, int voteCount) {
     this.title = title;
     this.lat = lat;
     this.lng = lng;
@@ -22,7 +22,7 @@ public class Location {
     this.keyString = null;
   }
 
-  public Location(String title, double lat, double lng, String note, long voteCount, String keyString) {
+  public Location(String title, double lat, double lng, String note, int voteCount, String keyString) {
     this.title = title;
     this.lat = lat;
     this.lng = lng;
@@ -52,7 +52,7 @@ public class Location {
     return note;
   }
 
-  public long getVoteCount() {
+  public int getVoteCount() {
     return voteCount;
   }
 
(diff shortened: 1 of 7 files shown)


GoMeet/src/main/java/com/google/sps/servlets/UpdateLocationServlet.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.jsoup.Jsoup;
+import org.jsoup.safety.Whitelist;
+
+/** Handles fetching and saving location data. */
+@WebServlet("/update-location-data")
+public class UpdateLocationServlet extends HttpServlet {
```
anna-giang, line 25: Optional suggestion: since this servlet will only update the vote (in fact, it will only increment the vote by 1), maybe consider naming it with a more specific name (e.g. VoteLocationServlet), unless you are considering to extend the functionality of this servlet in the future so that it will update more things than just increment the votes.

anna-giang:
I noticed that this PR contains commits from PR27, so I only looked at the code in the new commits (from 9a8a6c6 onwards) 😊 - I think it would be cleanest to first merge PR27, then rebase this branch on to main before merging, so that the duplicate commits don't affect anything.

anna-giang:
Looks good! 😊

GoMeet/src/test/java/com/google/sps/UpdateLocationServletTest.java
```diff
@@ -0,0 +1,93 @@
...
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  /** Tests if the entity's voteCount is updated. */
+  @Test
+  public void doPostTest() {
+    // Set up the database
+    DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+    Key entityKey = KeyFactory.createKey("Location", LOCATION_A.getTitle());
```
vivianho18, line 70: What is this line used for?

vivianho18 (commented):
Looks good! :) (+1 Anna's comment about merging then rebasing to avoid duplicate commits)

GoMeet/src/test/java/com/google/sps/UpdateLocationServletTest.java
theresenieva: This code has been replaced by code in the branch LocationDAOServer. :)

theresenieva:
I think the repeated commits will be removed from this PR once the previous PR is merged into master. I'm pretty sure they are appearing on this PR because it is part of the diff between ```main``` and ```LocationVote```. You can use EDIT to change the base branch which will remove the repeated commits (and then switch back).

GoMeet/src/main/java/com/google/sps/servlets/UpdateLocationServlet.java
theresenieva: One of my goal was to delete locations when vote count reaches zero, I think that will be handled in this servlet too.

GoMeet/src/main/java/com/google/sps/data/Location.java
```diff
@@ -8,15 +8,15 @@
   private final double lat;
   private final double lng;
   private final String note;
-  private int voteCount;
+  private long voteCount;
```
BipBopBoop: Why is this a long? I'm assuming that you'd never maximum integer value.

GoMeet/src/main/webapp/test/spec/spec-interactive-map.js
```diff
@@ -3,7 +3,7 @@ describe('Create map', function() {
   it ('Should create and return a map object', function() {
     const mapConstructorSpy = spyOn(google.maps, 'Map');
     const mockedMap = jasmine.createSpyObj('Map', ['addListener']);
-    mapConstructorSpy.and.returnValue(mockedMap);
+    mapConstructorSpy.and.returnValue(mockedMap);
```
BipBopBoop: nit: remove whitespace at the end of line.

GoMeet/src/main/webapp/test/spec/spec-interactive-map.js
```diff
@@ -93,6 +95,39 @@ describe('Post Location', function() {
...
+    // Set up mocks for Google MAPS API
+    const markerConstructorSpy = spyOn(google.maps, 'Marker');
+    const fakeMarker = jasmine.createSpyObj('Marker', ['addListener']);
+    markerConstructorSpy.and.returnValue(fakeMarker);
+    const infowindowConstructorSpy = spyOn(google.maps, 'InfoWindow');
+    let fakeMap = {};
+
+    await fetchLocations(fakeMap, mockedFetchWrapper);
+    expect(google.maps.Marker).toHaveBeenCalledWith(
+        {position: {lat: 10, lng: 15}, map: fakeMap});
```
BipBopBoop, line 159: Should we also verify that the title and note were retrieved appropriately as well?

BipBopBoop (commented):
Looks good, just a couple questions/nits

GoMeet/src/main/java/com/google/sps/data/Location.java
theresenieva: It was long because datastore stores the integer values as longs and getProperty("voteCount") would return a long. But, now that I think about it, I think it makes more sense to modify the data received from Datastore to suit my code than to modify my code to suit Datastore haha. I've changed it to int.

GoMeet/src/main/webapp/test/spec/spec-interactive-map.js
```diff
@@ -140,12 +140,34 @@ describe ('Fetch Locations', function() {
...
+    let note = false;
+    for (let i = 0; i < childNodes.length; i++) {
+      if (childNodes[i].textContent) {
+        let text = childNodes[i].textContent;
+        if (text === 'Hello') {
+          title = true;
+        } else if (text === 'My Note') {
+          note = true;
+        }
+      }
```
BipBopBoop: This isn't really ideal, one example that breaks this is if we were to use the same string for both the name and note. Do we not know which child node should have which text? Or maybe we could consider adding ids/classes to the elements so that they're better identifiable.

GoMeet/src/main/java/com/google/sps/data/Location.java
BipBopBoop: Great, thanks for explaining. That sounds like a reasonable change to me.

BipBopBoop (commented):
Thanks for the changes, added a comment on your test. Also, regarding your concern that we're leaning too heavily on buildInfoWindowInput(). I think it's okay to assume that this works, since it's being tested earlier. However, it brings up a good point that unit tests should really be testing one specific unit/function at a time. In our case this is testing the data retrieval functionality, so its appropriate to test only that, and mock everything else (presuming that those are being tested elsewhere). Either approach is reasonable.

GoMeet/src/main/webapp/test/spec/spec-interactive-map.js
theresenieva: I completely missed that - thanks for pointing it out.

PR #30, opened 2021-01-29
MeetingTime servlet

anna-giang:
Added a servlet at the `/meeting-time` endpoint to fetch existing and store new MeetingTime entities to Datastore.


GoMeet/src/main/java/com/google/sps/data/MeetingTimeEntityFields.java
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/** Names of fields of data in the Datastore MeetingTime entity */
+public class MeetingTimeFields {
```
vivianho18: This file should be renamed as MeetingTimeFields.java to match the class name

GoMeet/src/main/java/com/google/sps/servlets/MeetingTimeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+    Entity meetingTime = new Entity("MeetingTime");
+    Key meetingTimeKey = meetingTime.getKey();
+    meetingTime.setProperty(MeetingTimeFields.MEETING_TIME_ID, meetingTimeKey.toString());
+    meetingTime.setProperty(MeetingTimeFields.DATETIME, datetime);
+    meetingTime.setProperty(MeetingTimeFields.VOTE_COUNT, 0); // initially votes are 0
+    datastore.put(meetingTime);
+
+    // return the key of the created entity
+    response.setContentType("application/json");
+    response.getWriter().println(meetingTimeKey.toString());
```
theresenieva: Consider using KeyFactory.keyToString() instead, which converts the Key into a web-safe string. key.toString() is more suitable for debugging and logging. 

https://cloud.google.com/appengine/docs/standard/java/datastore/entities

theresenieva (approved):
One suggestion :)

GoMeet/src/main/java/com/google/sps/data/ErrorMessages.java
```diff
@@ -0,0 +1,27 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/** Names of fields of data in the Datastore MeetingTime entity */
```
vivianho18: I think this comment was copied over from the MeetingTimeEntityFields file!

anna-giang: Fixed! 😅

GoMeet/src/main/java/com/google/sps/data/MeetingTimeEntityFields.java
anna-giang: Yep, realised this just now 😅

GoMeet/src/main/java/com/google/sps/servlets/MeetingTimeServlet.java
anna-giang: Oh, okay, didn't know this. Thanks for pointing it out! 😊 Fixing it right now!

anna-giang:
Rebasing on main

anna-giang:
Changes in commit 3eacb85 was overwritten by the cherry-picked commits, so I moved it forward.

vivianho18:
LGTM! Might have been better to split commit [cac4806](https://github.com/googleinterns/step271-2020/pull/30/commits/cac4806bb26ec396abc8d109fc39864b78ef09f8) into 3 commits to reflect the 3 separate changes made as described in the commit message.

anna-giang:
> LGTM! Might have been better to split commit [cac4806](https://github.com/googleinterns/step271-2020/pull/30/commits/cac4806bb26ec396abc8d109fc39864b78ef09f8) into 3 commits to reflect the 3 separate changes made as described in the commit message.

Yep 😅 I was making the changes as I found the bugs without committing anything, and by the end of the whole debugging process, I realised it would be quite time consuming to back track and split up my changes into separate commits.

GoMeet/src/main/java/com/google/sps/servlets/MeetingTimeServlet.java
```diff
@@ -0,0 +1,118 @@
...
+    if (datetime == null) {
+      ServletUtil.sendErrorResponse(response, HttpServletResponse.SC_BAD_REQUEST, ErrorMessages.BAD_REQUEST_ERROR);
+      return;
+    }
+
+    Entity meetingTime = new Entity("MeetingTime");
+    meetingTime.setProperty(MeetingTimeFields.DATETIME, datetime);
+    meetingTime.setProperty(MeetingTimeFields.VOTE_COUNT, 0); // initially votes are 0
+    datastore.put(meetingTime);
+    String meetingTimeKey = KeyFactory.keyToString(meetingTime.getKey());
```
suprsrsly, line 108: Just a question - what do you do about meeting times that have been created but might be orphaned because the user session was interrupted between when the meeting time was created and when it was attached to the meeting?

suprsrsly: (Is that a valid question)

anna-giang: Hmm this is a valid question - I've googled around and potentially, we could associate the meeting time as a child entity of the meeting entity, and then routinely (e.g. every time the app launches) loop through every MeetingTime, checking whether their parent meeting exists, and delete them if the parent doesn't exist. 
There doesn't seem to be a way to easily cascade delete orphaned entities, it would need to be done by looping through and checking them all.

GoMeet/src/test/java/com/google/sps/MeetingTimeServletTest.java
```diff
@@ -98,6 +107,94 @@ public void testNullDatetimeQuery() throws IOException {
...
+    fakeMeetingTime1.setProperty(MeetingTimeFields.MEETING_TIME_ID, MEETING_TIME_ID_VAL);
+    datastore.put(fakeMeetingTime1); // stores to local datastore
+    Entity fakeMeetingTime2 = new Entity("MeetingTime");
+    fakeMeetingTime2.setProperty(MeetingTimeFields.MEETING_TIME_ID, MEETING_TIME_ID_VAL);
+    datastore.put(fakeMeetingTime2);
+
+    // fetch the entity
+    when(mockedRequest.getParameter(MeetingTimeFields.MEETING_TIME_ID)).thenReturn(MEETING_TIME_ID_VAL); 
+    new MeetingTimeServlet().doGet(mockedRequest, mockedResponse);
+<<<<<<< HEAD
```
suprsrsly: hmm...

GoMeet/src/main/java/com/google/sps/servlets/MeetingTimeServlet.java
suprsrsly: One possibility could be to adjust the state flow so that you would associate a meeting with every meeting time when the meeting time is created, but that might require some reworking to pass a meeting identifier around.

suprsrsly (approved):

PR #35, opened 2021-02-03
Email servlet

vivianho18:
Servlet to send emails to guests to invite them to a meeting event


GoMeet/src/main/java/com/google/sps/servlets/EmailServlet.java
```diff
@@ -0,0 +1,79 @@
...
+
+    try {
+      Message msg = new MimeMessage(session);
+      msg.setFrom(new InternetAddress("interns@go-meet.appspotmail.com"));
+      msg.addRecipient(Message.RecipientType.TO, new InternetAddress(emailAddress));
+      msg.setSubject("You have been invited to a GoMeet event!");
+      msg.setText("The meeting event ID is: " + meetingEventId); //TO DO: Find out what the meeting link should be
+      Transport.send(msg);
+    } catch (AddressException e) {
+      System.out.println("Address exception");
```
anna-giang: Optional: Since this and Line 74 are error messages, you may want to print it to standard error (`System.err.println()`), to keep the normal output and error output streams separate.

anna-giang: But this won't make a huge difference in out case, since we don't really use standard output for anything, hence optional 😊

GoMeet/src/test/java/com/google/sps/EmailServletTest.java
```diff
@@ -0,0 +1,108 @@
+package com.google.sps;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Mockito.*;
+import org.mockito.MockedStatic;
+import org.mockito.Mockito;
+import static org.mockito.Mockito.atLeast;
```
anna-giang: Nit: [Style guide](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing) says to list static and non-static imports in two blocks separated by one line, with static imports first. Within the blocks, imports are ordered alphabetically.

GoMeet/src/test/java/com/google/sps/EmailServletTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @Test 
+  public void testNullGuestList() throws IOException {
+    when(mockedRequest.getParameter("guestList")).thenReturn(null); 
+    new EmailServlet().doPost(mockedRequest, mockedResponse); 
+    testBadRequest(HttpServletResponse.SC_BAD_REQUEST, ErrorMessages.BAD_EMAIL_REQUEST_ERROR);
+  }
+
+  private void testBadRequest(int status, String errorMessage) throws IOException {
+    // Check that error response was sent with appropriate details
```
anna-giang, line 103: As you mentioned in our meeting, this is useful general function that maybe can be move to `ServletUtil`, because I can refactor my tests to use it too 😊
But the method signature would need to be changed to:
`private void testBadRequest(int status, String errorMessage, HttpServletResponse mockedResponse, StringWriter stringWriter)` 
--- do you think this parameter list is too long?

vivianho18: Yeh that's what I was unsure about too. But I think it makes more sense to have a long parameter list than to repeat the same helper method in all of our test files. I'll see what the hosts says when I asked them to review it after you and Therese have had a look :)

vivianho18: Also not sure if it should be included in ServletUtil or a separate class called ServletTestUtil

theresenieva: I think it would be better to include in the test folder, so ServletTestUtil :))

anna-giang: Hmm that is true... I think a new class ServletTestUtil, since it has nothing to do with actual servlet functionality.

GoMeet/src/test/java/com/google/sps/EmailServletTest.java
```diff
@@ -0,0 +1,108 @@
...
+      mockedServlet.verify(times(1), () -> EmailServlet.sendEmail(GUEST_2, MEETING_EVENT_ID));
+      mockedServlet.verify(times(1), () -> EmailServlet.sendEmail(GUEST_3, MEETING_EVENT_ID));
+    }
+
+    // Assert that key is returned in JSON Object
+    HashMap<String, Object> keyObj = new HashMap<String, Object>() {{
+      put("sent", true);
+    }};
+    writer.flush(); // writer may not have been flushed yet
+    assertTrue(stringWriter.toString().contains(ServletUtil.convertMapToJson(keyObj)));
```
theresenieva: One comment I got on my prev PR, maybe it would be better to convert the stringWriter.toString() back to a hashmap and then test whether the value of "sent" is as expected. 

The comment:
https://github.com/googleinterns/step271-2020/pull/27#discussion_r567582833 

I think the reasoning would still apply even though that test was using Gson.toJson() and this is JSONObject.toString().

theresenieva (approved):
Few suggestions. Other than that, lgtm :))

GoMeet/src/test/java/com/google/sps/EmailServletTest.java
vivianho18: That's a good point! I've updated these tests, and have made a note to do the same for the other test classes

GoMeet/src/main/java/com/google/sps/data/ServletUtil.java
```diff
@@ -72,16 +72,17 @@ public static void sendErrorResponse(HttpServletResponse response, int status, S
    * @return the decoded URI
    */
   public static String decodeUri(String encodedUri) {
+    String result; 
     // Decode "%2C" to ","
-    encodedUri.replace("%2C", ",");
```
ghost: Consider using [URLDecoder](https://docs.oracle.com/javase/7/docs/api/java/net/URLDecoder.html) to do this instead.

GoMeet/src/main/java/com/google/sps/servlets/EmailServlet.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    String guestList = ServletUtil.decodeUri(guestListUri); 
+    String[] guestListSplit = guestList.split(",");
+    for (int i = 0; i < guestListSplit.length; i++) {
+      boolean sent = sendEmail(guestListSplit[i], meetingEventId);
+      if (!sent) {
+        ServletUtil.sendErrorResponse(response, HttpServletResponse.SC_BAD_REQUEST, 
+            ErrorMessages.BAD_EMAIL_REQUEST_ERROR);
+        return;
```
ghost: Would it be better to keep trying to send it to the remaining emails in the list instead of returning immediately?

GoMeet/src/main/java/com/google/sps/data/ServletUtil.java
BipBopBoop: +1, it's generally better to use preexisting tools, but now that you've already made your own iteration, it's not necessary.

GoMeet/src/main/java/com/google/sps/servlets/EmailServlet.java
BipBopBoop: This is a good point. Either approach is reasonable, but we should at least add a comment here, or in the design, as to why we chose how to handle this error.

GoMeet/src/test/java/com/google/sps/EmailServletTest.java
BipBopBoop: Reusing code and avoiding duplication is great, if you think this can be reused in other tests, then it makes sense to put it somewhere where the other tests can access it.

BipBopBoop (commented):
Generally LGTM. Mostly just nits, but I would like a better understanding of your decision behind error handling before I +1.

GoMeet/src/main/java/com/google/sps/data/ServletUtil.java
vivianho18: Ok I'll keep this in mind for next time

vivianho18:
> Generally LGTM. Mostly just nits, but I would like a better understanding of your decision behind error handling before I +1.

GoMeet/src/test/java/com/google/sps/EmailServletTest.java
vivianho18: Yep, this function has been added to a ServletTestUtil class in [PR#40](https://github.com/googleinterns/step271-2020/pull/40) for Anna and Therese to use too :)

GoMeet/src/main/java/com/google/sps/servlets/EmailServlet.java
vivianho18: I have changed the design to continue calling the send email function - the json response is now a hashmap with each email address as a key and the corresponding sent status (true/false) as the value. I will probably use this response data to display a message on the html page e.g. "Emails sent successfully to X, Y and Z. Unable to send emails to A and B."

GoMeet/src/test/java/com/google/sps/EmailServletTest.java
```diff
@@ -67,19 +67,22 @@ public void testSuccessfulDoPost() throws IOException {
...
     // Convert expected and result to Json Object to compare 
     String resultsJsonStr = stringWriter.toString(); 
-    String expectedJsonStr = ServletUtil.convertMapToJson(keyObj); 
+    String expectedJsonStr = ServletUtil.convertMapToJson(expected); 
     JsonObject resultsJsonObj = new JsonParser().parse(resultsJsonStr).getAsJsonObject();
     JsonObject expectedJsonObj = new JsonParser().parse(expectedJsonStr).getAsJsonObject(); 
-    assertEquals(resultsJsonObj.get("sent"), expectedJsonObj.get("sent"));
+    assertEquals(resultsJsonObj.get(GUEST_1), expectedJsonObj.get(GUEST_1));
+    assertEquals(resultsJsonObj.get(GUEST_1), expectedJsonObj.get(GUEST_1));
+    assertEquals(resultsJsonObj.get(GUEST_1), expectedJsonObj.get(GUEST_1));
```
BipBopBoop: nit: I'm assuming you meant to increment the GUEST_* here.

GoMeet/src/main/java/com/google/sps/servlets/EmailServlet.java
BipBopBoop: Sounds good to me :)

BipBopBoop (approved):
LGTM, just a nit. Thanks for making the changes :)

GoMeet/src/test/java/com/google/sps/EmailServletTest.java
vivianho18: Yepp my bad

PR #36, opened 2021-02-03
Location dao server

theresenieva:
Create a LocationDAO class. 

I chose to do this to: (1) Avoid duplicated code, (2) Make the code more testable as I can mock the locationDao object and remove the database dependency from the servlet tests, (3) If the structure of the database changes, then the code will be more convenient to update as the changes only need to be made to the LocationDao.java.

### DIFF
--- a/GoMeet/src/main/java/com/google/sps/data/Dao.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Dao.java
@@ -10,7 +10,7 @@ public interface Dao<T> {
 
   List<T> getAll();
 
-  void save(T t);
+  String save(T t);
 
   void update(String keyString);
 
--- a/GoMeet/src/main/java/com/google/sps/data/Dao.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Dao.java
@@ -1,5 +1,7 @@
 package main.java.com.google.sps.data;
 
+import com.google.appengine.api.datastore.EntityNotFoundException;
+
 import java.util.List;
 import java.util.Optional;
 
@@ -12,7 +14,7 @@ public interface Dao<T> {
 
   String save(T t);
 
-  void update(String keyString);
+  void update(String keyString) throws EntityNotFoundException;
 
   void delete(String keyString);  
 }
--- a/GoMeet/src/main/java/com/google/sps/data/Dao.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Dao.java
@@ -14,7 +14,7 @@ public interface Dao<T> {
 
   String save(T t);
 
-  void update(String keyString) throws EntityNotFoundException;
+  void updateVote(String keyString) throws EntityNotFoundException;
 
   void delete(String keyString);  
 }
--- a/GoMeet/src/main/java/com/google/sps/data/Dao.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Dao.java
@@ -1,0 +1,18 @@
+package main.java.com.google.sps.data;
+
+import java.util.List;
+import java.util.Optional;
+
+public interface Dao<T> {
+  // TODO: Visibility
+
+  Optional<T> get(String keyString);
+
+  List<T> getAll();
+
+  void save(T t);
+
+  void update(String keyString);
+
+  void delete(String keyString);  
+}
--- a/GoMeet/src/main/java/com/google/sps/data/Location.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Location.java
@@ -63,4 +63,22 @@ public class Location {
   public String getKeyString() {
     return keyString;
   }
+
+  @Override
+  public boolean equals(Object o) {
+    if (o == this) {
+      return true;
+    }
+    if (!(o instanceof Location)) {
+      return false;
+    }
+    Location otherLocation = (Location) o;
+    
+    return this.title.equals(otherLocation.getTitle()) &&
+        (Double.compare(this.lat, otherLocation.getLat()) == 0) &&
+        (Double.compare(this.lng, otherLocation.getLng()) == 0) &&
+        this.note.equals(otherLocation.getNote()) &&
+        ((this.keyString == otherLocation.getKeyString()) ||
+            this.keyString.equals(otherLocation.getKeyString()));
+  }
 }
--- a/GoMeet/src/main/java/com/google/sps/data/Location.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Location.java
@@ -74,6 +74,7 @@ public class Location {
     }
     Location otherLocation = (Location) o;
     
+    // TODO: Situation when one key string is null
     return this.title.equals(otherLocation.getTitle()) &&
         (Double.compare(this.lat, otherLocation.getLat()) == 0) &&
         (Double.compare(this.lng, otherLocation.getLng()) == 0) &&
--- a/GoMeet/src/main/java/com/google/sps/data/Location.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Location.java
@@ -82,7 +82,6 @@ public class Location {
     }
     Location otherLocation = (Location) o;
     
-    // TODO: Situation when one key string is null
     return this.title.equals(otherLocation.getTitle()) &&
         (Double.compare(this.lat, otherLocation.getLat()) == 0) &&
         (Double.compare(this.lng, otherLocation.getLng()) == 0) &&
--- a/GoMeet/src/main/java/com/google/sps/data/Location.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Location.java
@@ -64,6 +64,11 @@ public class Location {
     return keyString;
   }
 
+  /** 
+   * Checks if this location is equal to another location. 
+   * Locations are equal if they have the same title, lat, lng, note and key string.
+   * @param o the object to compare this location to.
+   */
   @Override
   public boolean equals(Object o) {
     if (o == this) {
--- a/GoMeet/src/main/java/com/google/sps/data/Location.java
+++ b/GoMeet/src/main/java/com/google/sps/data/Location.java
@@ -66,7 +66,10 @@ public class Location {
 
   /** 
    * Checks if this location is equal to another location. 
-   * Locations are equal if they have the same title, lat, lng, note and key string.
+   * Locations are equal if they have the same title, lat, lng, note and the 
+   * key strings are not different.
+   * Note: Location entities are the same even if one key string is null.
+   * That is because locations' MeetingID+Title are unique.
    * @param o the object to compare this location to.
    */
   @Override
@@ -85,6 +88,8 @@ public class Location {
         (Double.compare(this.lng, otherLocation.getLng()) == 0) &&
         this.note.equals(otherLocation.getNote()) &&
         ((this.keyString == otherLocation.getKeyString()) ||
-            this.keyString.equals(otherLocation.getKeyString()));
+        this.keyString == null ||
+        otherLocation.getKeyString() == null ||
+        this.keyString.equals(otherLocation.getKeyString()));
   }
 }
--- a/GoMeet/src/main/java/com/google/sps/data/LocationDao.java
+++ b/GoMeet/src/main/java/com/google/sps/data/LocationDao.java
@@ -50,9 +50,21 @@ public class LocationDao implements Dao<Location> {
     return locations;
   }
 
+  /** 
+   * Stores a new location on datastore.
+   * Returns the key string of the new entity.
+   */
   @Override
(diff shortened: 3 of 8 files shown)


GoMeet/src/main/java/com/google/sps/data/Dao.java
```diff
@@ -0,0 +1,20 @@
+package main.java.com.google.sps.data;
+
+import com.google.appengine.api.datastore.EntityNotFoundException;
+
+import java.util.List;
+import java.util.Optional;
+
+public interface Dao<T> {
```
vivianho18, line 8: Is this going to be a general class that we will all use? If not, it might be better to name it something more specific to the location feature

GoMeet/src/main/java/com/google/sps/data/Location.java
```diff
@@ -63,4 +63,23 @@ public int getVoteCount() {
   public String getKeyString() {
     return keyString;
   }
+
+  @Override
+  public boolean equals(Object o) {
```
vivianho18, line 76: Not entirely sure what this function does - would be helpful to add a comment describing the functionality and how/when it would be used

GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -1,38 +1,37 @@
...
  /** Responds with a JSON array containing location data. */
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     response.setContentType("application/json");
 
-    Collection<Location> locations = getLocations();
-    Gson gson = new Gson();
-    String json = gson.toJson(locations);
+    List<Location> locations = locationDao.getAll();
+    String json = ServletUtil.convertToJson(locations);
```
vivianho18, line 34: Double check if this is actually converting to JSON properly. I know Anna and I ran into issues with calling this function and it was returning null instead of the JSON string.

GoMeet/src/main/java/com/google/sps/servlets/UpdateLocationServlet.java
```diff
@@ -23,25 +22,22 @@
...
+  private Dao<Location> locationDao = new LocationDao();
+
   /** Accepts a POST request with information about the location entity to add a vote to. */
   @Override
-  public void doPost(HttpServletRequest request, HttpServletResponse response) {
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String keyString = request.getParameter("key");
     try {
-        updateLocation(keyString);
+      locationDao.update(keyString);
```
vivianho18: Would be good to send something in the response to verify that the request was successful

GoMeet/src/test/java/com/google/sps/LocationDaoTest.java
```diff
@@ -0,0 +1,171 @@
...
+@RunWith(JUnit4.class)
+public class LocationDaoTest {
+  private final long INIT_VOTE_COUNT = 1;
+  private final Location LOCATION_A = new Location("Cake Shop", 10.0, 15.0, "I like Cakes!", 1);
+  private final Location LOCATION_B = new Location("Fruit Shop", 22.0, 60.0, "", 1);
+
+  private final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+  private DatastoreService ds;
+  private LocationDao ld;
```
vivianho18: nit - rename variable to locationDao so that its more obvious

GoMeet/src/test/java/com/google/sps/LocationDaoTest.java
```diff
@@ -0,0 +1,171 @@
...
+   */
+  @Test 
+  public void getAllTestMultiple() {
+    // Set up database
+    String key1 = addLocationToDatabase(LOCATION_A);
+    String key2 = addLocationToDatabase(LOCATION_B);
+
+    List<Location> result = ld.getAll(); 
+    
+    assertEquals(2, result.size());
```
vivianho18: Would also be good to check the contents of the result list i.e. 
```
assertEquals(LOCATION_A, result.get(0))
assertEquals(LOCATION_B, result.get(1))
```

GoMeet/src/test/java/com/google/sps/LocationDaoTest.java
```diff
@@ -0,0 +1,171 @@
...
+    location.setProperty("note", LOCATION_A.getNote());
+    location.setProperty("voteCount", LOCATION_A.getVoteCount());
+    ds.put(location);
+
+    String keyString = KeyFactory.keyToString(location.getKey());
+
+    try {
+      ld.update(keyString);
+      Entity retrievedLocation = ds.get(location.getKey());
+      assertEquals(2, ((Long) retrievedLocation.getProperty("voteCount")).intValue());
```
vivianho18, line 152: nit - not sure if its really necessary, but could also check the other properties to ensure they haven't been modified/lost

vivianho18 (commented):
Good job on the refactoring! LGTM :)

GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
theresenieva: Yup, will do. Thanks for letting me know.

GoMeet/src/main/java/com/google/sps/data/LocationDao.java
```diff
@@ -0,0 +1,87 @@
...
+    entity.setProperty("note", location.getNote());
+    entity.setProperty("voteCount", location.getVoteCount());
+    ds.put(entity);
+
+    return KeyFactory.keyToString(entity.getKey());
+  }
+
+  /** Increments the vote count of the location entity with the give keyString. */
+  @Override 
+  public void update(String keyString) throws EntityNotFoundException {
```
anna-giang: Since this function only updates the vote count, would a more specific name (e.g. updateVote, or even increaseVote) be appropriate? Unless you're thinking of add more functionality to this function in the future so that is updates more than just the vote count.

GoMeet/src/main/java/com/google/sps/data/Dao.java
theresenieva: I think it would be best to keep the name non-specific (so Dao instead of LocationDao) because it is a generic interface. For your question, we don't all have to use it. It is up to you if you would like to use it and what you think is best for your feature (I found that this was useful for me as a I had multiple servlets accessing the same data).

GoMeet/src/main/java/com/google/sps/servlets/UpdateLocationServlet.java
theresenieva: Yep, sounds like a good idea.

GoMeet/src/test/java/com/google/sps/LocationDaoTest.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.After;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import static org.mockito.Mockito.*;
```
anna-giang: Nit: Static imports should be listed first in a separate block from non-static import, according to [style guide](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing).

GoMeet/src/test/java/com/google/sps/LocationServletTest.java
```diff
@@ -75,51 +78,39 @@ public void tearDown() {
...
+  @Test
+  public void doPostTest() throws IOException {
+    // Set up request mock
+    when(request.getParameter("title")).thenReturn(LOCATION_A.getTitle());
+    when(request.getParameter("lat")).thenReturn(Double.toString(LOCATION_A.getLat()));
+    when(request.getParameter("lng")).thenReturn(Double.toString(LOCATION_A.getLng()));
+    when(request.getParameter("note")).thenReturn(LOCATION_A.getNote());
+
+    // Set up response mock writer
     StringWriter stringWriter = new StringWriter();
```
anna-giang: Could potentially define `stringWriter` and `writer` as private instance variables, so you don't have to set up the response mock writer in both test cases separately.

theresenieva: Yup, sounds good. Will do

GoMeet/src/main/java/com/google/sps/data/Dao.java
vivianho18: Ok that makes sense, agreed I think this was a good style choice for your implementation :)

GoMeet/src/test/java/com/google/sps/LocationDaoTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+  private final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+  private DatastoreService ds;
+  private LocationDao locationDao;
+
+  @Before
+  public void setUp() {
+    helper.setUp();
+    ds = DatastoreServiceFactory.getDatastoreService();
```
suprsrsly, line 56: Does this create a real connection to the datastore service or is this faked?

theresenieva: It is faked. It is a local implementation of Datastore.

https://cloud.google.com/appengine/docs/standard/java/tools/localunittesting

GoMeet/src/test/java/com/google/sps/LocationServletTest.java
```diff
@@ -75,51 +76,35 @@ public void tearDown() {
...
-
-    // Check the entity property values were assigned correctly
-    assertEquals(TITLE_A, result.getProperty("title"));
-    assertEquals(LAT_A_VALUE, result.getProperty("lat"));
-    assertEquals(LNG_A_VALUE, result.getProperty("lng"));
-    assertEquals(NOTE_A, result.getProperty("note"));
-    assertEquals(INIT_VOTE_COUNT, ((Long) result.getProperty("voteCount")).intValue());
+    // Check the mockedDao was called with the correct parameters
+    ArgumentCaptor<Location> captor = ArgumentCaptor.forClass(Location.class);
+    verify(mockedLocationDao, times(1)).save(captor.capture());
```
suprsrsly, line 100: Just to confirm, your test is checking to see if LocationDao::save() is called once with the argument LOCATION_A?

GoMeet/src/test/java/com/google/sps/UpdateLocationServletTest.java
```diff
@@ -46,48 +56,78 @@
...
 
-    String keyString = KeyFactory.keyToString(location.getKey());
+    UpdateLocationServlet servlet = new UpdateLocationServlet();
+    servlet.setDao(mockedLocationDao);
+    servlet.doPost(request, response);
+    
+    try {
+      verify(mockedLocationDao, times(1)).updateVote(keyString);
+      verify(response, times(1)).setStatus(HttpServletResponse.SC_OK);
+    } catch (EntityNotFoundException e) {
```
suprsrsly: Why are you wrapping this in try/catch?

suprsrsly: also, if try/catch is necessary, it's probably not necessary to only catch EntityNotFoundException s.

GoMeet/src/test/java/com/google/sps/LocationServletTest.java
theresenieva: Yes, that is correct. And then the next assertEquals is to check that the response was sent with the keyString.

GoMeet/src/test/java/com/google/sps/UpdateLocationServletTest.java
theresenieva: Because updateVote() throws an EntityNotFoundException. Would it be better to throw the exception up?

suprsrsly: It seems like the test is meant to fail if an exception is thrown because in the catch(){} block, there is a single fail() call.

It seems like only catching EntityNotFoundException makes the test somewhat fragile because if any other exception is thrown, the test could pass, but it seems to me like you want the test to fail if there is any failure in any of the calls, since this is a test of a "working" code path.

So, I'm wondering why you wrap the important parts of the test in a try/catch call instead of letting the (or any) exception bubble up into the test framework which I suppose should cause the test to fail anyway.

suprsrsly (approved):

PR #40, opened 2021-02-05
Servlet test util

vivianho18:
Create a shared ServletTestUtil class to hold the helper function testBadRequest and refactor tests

### DIFF
--- a/GoMeet/src/test/java/com/google/sps/ServletTestUtil.java
+++ b/GoMeet/src/test/java/com/google/sps/ServletTestUtil.java
@@ -1,0 +1,48 @@
+package test.java.com.google.sps;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.*;
+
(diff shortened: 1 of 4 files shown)


theresenieva (approved):
Looks fine :)

GoMeet/src/test/java/com/google/sps/ServletTestUtil.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Class for servlet unit testing helper functions 
+ */
+public class ServletTestUtil {
+  
+  /**
+   * Helper function to test the correct error response was sent in the case of 
+   * a bad request
+   */
+  public static void badRequest(int status, String errorMessage,
```
suprsrsly: I would call this function "expectBadRequest" or similar because it has some assertions that are meant to pass inside it.

suprsrsly (approved):
Great cleanups!

PR #32, opened 2021-02-01
Vote Meeting Times JavaScript DAO

anna-giang:
This DAO will be used to access the servlet at `/vote-meeting-time` endpoint to increment the votes and add to the voters list of the meeting time voted for.

PR #41, opened 2021-02-05
Email DAO

vivianho18:
Data access object for sending a post request to the email servlet at endpoint `/email`

### DIFF
--- a/GoMeet/src/main/webapp/js/email-dao.js
+++ b/GoMeet/src/main/webapp/js/email-dao.js
@@ -1,0 +1,35 @@
+class EmailDAO {
+  /** Data Access Object for sending emails */
+
+  static endpoint = '/email'; // URL endpoint of the associated servlet
+  
+  /**
+   * Completes a POST request to send an email invitation with the 
+   * meeting event link to all addresses in the guest list 
+   * @param {String} meetingTimeId the unique entity key of the 
+   * MeetingEvent that the guest list is invited to 
+   * @param {List} guestList the list of guests' email addresses 
+   * @return JSON containing the sent status of the emails  
+   */
+  static async inviteGuestsToMeeting(meetingEventId, guestList) {
+    if (meetingEventId === null || meetingEventId === undefined) {
+      throw new Error(INSUFFICIENT_REQUEST_PARAM);
+    }
+
+    if (guestList === null || guestList === undefined) {
+      throw new Error(INSUFFICIENT_REQUEST_PARAM); 
+    }
+
+    if (typeof meetingEventId !== 'string' || typeof guestList !== 'object' || typeof guestList[1] !== 'string') {
+      throw new Error(INVALID_PARAM_TYPE); 
+    }
+
+    // Encode the data to be sent in the query string
+    let urlString = DAOUtils.url(EmailDAO.endpoint, {'meetingEventId': meetingEventId, 
+    'guestList': guestList});
+
+    let responseInit = {method: 'POST'}
+    let response = await fetch(urlString, responseInit).then((response) => response.json());
+    return response;
+  }
+}
--- a/GoMeet/src/main/webapp/js/email-dao.js
+++ b/GoMeet/src/main/webapp/js/email-dao.js
@@ -12,16 +12,19 @@ class EmailDAO {
    * @return JSON containing the sent status of the emails  
    */
   static async inviteGuestsToMeeting(meetingEventId, guestList) {
-    if (meetingEventId === null || meetingEventId === undefined) {
+    if (meetingEventId === null || meetingEventId === undefined ||
+        guestList === null || guestList === undefined) {
       throw new Error(INSUFFICIENT_REQUEST_PARAM);
     }
 
-    if (guestList === null || guestList === undefined) {
-      throw new Error(INSUFFICIENT_REQUEST_PARAM); 
+    if (typeof meetingEventId !== 'string' || typeof guestList !== 'object') {
+      throw new Error(INVALID_PARAM_TYPE); 
     }
 
-    if (typeof meetingEventId !== 'string' || typeof guestList !== 'object' || typeof guestList[1] !== 'string') {
-      throw new Error(INVALID_PARAM_TYPE); 
+    for (let i = 0; i < guestList.length; i++) {
+      if (typeof guestList[i] !== 'string') {
+        throw new Error(INVALID_PARAM_TYPE); 
+      }
     }
 
     // Encode the data to be sent in the query string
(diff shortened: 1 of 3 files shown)


GoMeet/src/main/webapp/js/email-dao.js
```diff
@@ -0,0 +1,35 @@
...
+   * MeetingEvent that the guest list is invited to 
+   * @param {List} guestList the list of guests' email addresses 
+   * @return JSON containing the sent status of the emails  
+   */
+  static async inviteGuestsToMeeting(meetingEventId, guestList) {
+    if (meetingEventId === null || meetingEventId === undefined) {
+      throw new Error(INSUFFICIENT_REQUEST_PARAM);
+    }
+
+    if (guestList === null || guestList === undefined) {
```
anna-giang: This if-block and the above if-block can be combined into one for brevity. This is what Daniel mentioned in my previous [PR](https://github.com/googleinterns/step271-2020/pull/32) 👍

GoMeet/src/main/webapp/js/email-dao.js
```diff
@@ -0,0 +1,35 @@
...
+      throw new Error(INSUFFICIENT_REQUEST_PARAM); 
+    }
+
+    if (typeof meetingEventId !== 'string' || typeof guestList !== 'object' || typeof guestList[1] !== 'string') {
```
anna-giang: Why is the code only checking the second element of the guestList?

vivianho18: Whoops this was meant to be guestList[0] but yes, I should check each element in the array. I was thinking it was a java array and assumed that each element would be of the same type.

theresenieva (approved):
Yep, lgtm

GoMeet/src/main/webapp/js/email-dao.js
```diff
@@ -0,0 +1,38 @@
...
+
+    for (let i = 0; i < guestList.length; i++) {
+      if (typeof guestList[i] !== 'string') {
+        throw new Error(INVALID_PARAM_TYPE); 
+      }
+    }
+
+    // Encode the data to be sent in the query string
+    let urlString = DAOUtils.url(EmailDAO.endpoint, {'meetingEventId': meetingEventId, 
+    'guestList': guestList});
```
ghost: nit: I'd format this something like this:

```
    let urlString = DAOUtils.url(
      EmailDAO.endpoint,
      {'meetingEventId': meetingEventId, 'guestList': guestList}
    );
```
Always indent line continuations otherwise it's not always obvious at a glance whether this is one line of code or two.

GoMeet/src/main/webapp/js/email-dao.js
```diff
@@ -0,0 +1,38 @@
...
+
+    let responseInit = {method: 'POST'}
```
ghost: semicolon

GoMeet/src/main/webapp/test/spec/spec-email-dao.js
```diff
@@ -0,0 +1,91 @@
...
+  const QUERY_STRING = '?meetingEventId=' + encodeURIComponent(MEETING_EVENT_ID) + 
+      '&guestList=' + encodeURIComponent(GUEST_LIST); 
+  
+  beforeEach(function() {
+    spyOn(window, 'fetch').and.callFake(async function(url) {
+      // Assume all fetches other than the hardcoded const QUERY_STRING 
+      // to be invalid
+      if (url !== EmailDAO.endpoint + QUERY_STRING) {
+        return new Response(JSON.stringify(ERROR_RESPONSE), null);
+      } else {
```
ghost: nit: I would omit the `else` here and move line 24 outside the if statement to make it super clear that this function always returns something.

GoMeet/src/main/webapp/js/email-dao.js
```diff
@@ -0,0 +1,38 @@
...
+   * @param {List} guestList the list of guests' email addresses 
+   * @return JSON containing the sent status of the emails  
+   */
+  static async inviteGuestsToMeeting(meetingEventId, guestList) {
+    if (meetingEventId === null || meetingEventId === undefined ||
+        guestList === null || guestList === undefined) {
+      throw new Error(INSUFFICIENT_REQUEST_PARAM);
+    }
+
+    if (typeof meetingEventId !== 'string' || typeof guestList !== 'object') {
```
anna-giang: Hmm maybe you should check whether guestList is an Array (you can do it like [this](https://stackoverflow.com/questions/4775722/how-to-check-if-an-object-is-an-array))? 
Because a lot of different types fall under 'object'.

GoMeet/src/main/webapp/test/spec/spec-email-dao.js
```diff
@@ -0,0 +1,99 @@
+describe ('EmailDAO - inviteGuestsToMeeting', function() {
+  const SENT_RESULT = ['sent', true]; 
+  const ERROR_RESPONSE = {
+    status: 404, 
+    messsage: 'Some error message'
+  };
+  const RESPONSE_INIT = {method: 'POST'};
+  const MEETING_EVENT_ID = 'qwerty12345'; 
+  // Note: Guest list should be of type object and each element within the list
```
anna-giang: 'Guest list should be type 'Array'?', following from the comment above.

ghost (approved):
LGTM! Small nits only.

GoMeet/src/main/webapp/test/spec/spec-email-dao.js
```diff
@@ -0,0 +1,99 @@
...
+        return new Response(JSON.stringify(ERROR_RESPONSE), null);
+      } else {
+        return new Response(JSON.stringify(SENT_RESULT), null);
+      }
+    }); 
+  }); 
+  
+  it('Throws an exception if the meetingEventId param is not provided', async function () {
+    let errorMessage;
+    try {
```
anna-giang: Optional: you could extract the test for the error responses to a method, so that the code is less repeated, like what I did in `spec-vote-meeting-time-dao.js` [here](https://github.com/googleinterns/step271-2020/blob/main/GoMeet/src/main/webapp/test/spec/spec-vote-meeting-time-dao.js)

anna-giang (approved):
Looks good to me, with optional changes.

GoMeet/src/main/webapp/js/email-dao.js
vivianho18: Good point

GoMeet/src/main/webapp/test/spec/spec-email-dao.js
vivianho18: Done thanks for the suggestion!

PR #42, opened 2021-02-07
Remove '?' from query string in MeetingTimeDAO tests, VoteMeetingTimeDAO

anna-giang:
This follows the refactoring of the DAOUtils.url function to add the '?' to the query string instead of including the '?' with each of the endpoints in the DAOs.


theresenieva (approved):
Yep, easy change. Looks good :)

vivianho18 (commented):
lgtm :)

suprsrsly (approved):
This is OK because the change is pretty small, but ideally it would be nice to wrap the generation of a request in a library so that you don't need to worry about moving around question mark characters. What happens in some hypothetical (but admittedly unlikely for the scope of this project) situation where you have hundreds of query strings scattered around that you need to track down manually?

PR #43, opened 2021-02-08
Display popular locations

theresenieva:
Display popular locations on the page. 

I've created a servlet to return a JSON array of the popular locations and JavaScript code to display those locations. I chose to display the popular locations as a list below the map when the user clicks a button. I think this would be cleaner than having the list update whenever a new location is added.


GoMeet/src/main/java/com/google/sps/servlets/PopularLocationServlet.java
```diff
@@ -0,0 +1,60 @@
...
+    String json = ServletUtil.convertToJson(popularLocations);
+    response.getWriter().println(json);
+  }
+
+  /** Returns a list of the locations with the highest vote count. */
+  private List<Location> getPopularLocation() {
+    List<Location> popularLocations = new ArrayList<>();
+    List<Location> allLocations = locationDao.getAll();
+
+    int maxVoteCount = 0;
```
theresenieva, line 42: I decided to iterate through the list of locations to find the most popular locations as it was:
(1) O(n) each location is visited once.
(2) Given time-constraints, this approach is simple to implement,
(3) In our design doc, we specified that each meeting will have max. 5 locations (so a small set of data).

I considered sorting the list first based on the vote count and then returning the first elements in the list but Collection.sort() has a complexity of O(n*log(n)).

In future, if this feature were to be scaled up, I think storing a list of popular locations as a property of the meeting entity would be more efficient (and updating it when a new location is added) as each location entity for that meeting does not need to be read when requesting the popular locations.

PR #45, opened 2021-02-08
Fixes to location display

theresenieva:
Make improvements to how location information is displayed on the map.

Previously, the page had to be refreshed to see the new vote count or new location. I've made changes so that the vote count and location are displayed immediately on the map.

### DIFF
--- a/GoMeet/src/main/webapp/js/interactive-map.js
+++ b/GoMeet/src/main/webapp/js/interactive-map.js
@@ -116,8 +116,8 @@ function buildInfoWindowVote(title, voteCount, note, keyString) {
   const button = document.createElement('button');
   button.setAttribute('id', 'voteButton');
   button.appendChild(document.createTextNode('VOTE'));
-  button.onclick = () => {
-    MeetingLocationDAO.updateLocation(keyString);
+  button.onclick = async () => {
+    await MeetingLocationDAO.updateLocation(keyString);
     const currVote = voteContainer.innerText;
     voteContainer.innerText = parseInt(currVote) + 1;
   };
--- a/GoMeet/src/main/webapp/js/interactive-map.js
+++ b/GoMeet/src/main/webapp/js/interactive-map.js
@@ -1,3 +1,5 @@
+const INITIAL_VOTE_COUNT = 1;
+
 /** Initialises the map. */
 function initMap() {
   // Create the map object
@@ -27,7 +29,7 @@ function createLocationForEdit(map, lat, lng) {
 
   const infoWindow =
       new google.maps.InfoWindow({content: buildInfoWindowInput(lat, lng,
-         editLocation)});
+         editLocation, map)});
 
   // When the user closes the editable info window, remove the marker.
   google.maps.event.addListener(infoWindow, 'closeclick', () => {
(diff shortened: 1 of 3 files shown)


GoMeet/src/main/webapp/js/meeting-location-dao.js
```diff
@@ -23,8 +23,8 @@ class MeetingLocationDAO {
     params.append('lat', lat);
     params.append('lng', lng);
     params.append('note', note);
-    let result = await fetch(this.storingEndPoint, {method: 'POST', body: params})
-      .then((response) => response.json());
+    let response = await fetch(this.storingEndPoint, {method: 'POST', body: params});
+    let result = await response.json();
```
vivianho18, line 44: Why did you change from arrow function?

vivianho18 (commented):
looks good :))

GoMeet/src/main/webapp/js/meeting-location-dao.js
theresenieva: I received a comment on a previous PR that using two await statements is cleaner. It is just a style choice.

vivianho18: Oh my bad! - I thought we were told to preferentially use arrow functions for these async calls

theresenieva: Haha nw!

GoMeet/src/main/webapp/js/interactive-map.js
```diff
@@ -110,24 +109,31 @@ async function fetchLocations(map) {
...
-  noteContainer.innerText = note;
+  const titleContainer = createSpanContainer(title, 'displayTitle');
+  const noteContainer = createSpanContainer(note, 'displayNote');
+  const voteContainer = createSpanContainer(voteCount, 'displayVoteCount');
 
   const button = document.createElement('button');
+  button.setAttribute('id', 'voteButton');
   button.appendChild(document.createTextNode('VOTE'));
   button.onclick = () => {
     MeetingLocationDAO.updateLocation(keyString);
```
anna-giang: Did you need an `async` keyword here?

GoMeet/src/main/webapp/js/interactive-map.js
```diff
@@ -110,24 +109,31 @@ async function fetchLocations(map) {
...
+    const currVote = voteContainer.innerText;
+    voteContainer.innerText = parseInt(currVote) + 1;
```
anna-giang, line 122: The code is directly updating the HTML display voteCount, which could be dangerous in that if the `MeetingLocationDAO.updateLocation()` call fails to update (error occurs during update), then the voteCount in the UI is inconsistent with the voteCount in the server. 

Will this be refactored later on to re-render the page to update the voteCount instead of directly updating it?

theresenieva: Yep, that's a good point. I do plan to refactor but instead I plan to make updateLocation() check the status of the response. If it is not 200 then updateLocation() will throw an error. In the button.onclick() it would look like this:

```
try {
    await MeetingLocationDAO.updateLocation(keyString);
    const currVote = voteContainer.innerText;
    voteContainer.innerText = parseInt(currVote) + 1;
} catch (error) {
  handleError();
}
```

That way the innerText is only updated if the updateLocation() does not throw an error. 

Does this seem like a good idea?

GoMeet/src/main/webapp/js/interactive-map.js
theresenieva: Ahhh, yup. I do. Thanks for pointing that out :)

GoMeet/src/main/webapp/js/interactive-map.js
anna-giang: Yep sounds good! I think its okay for this PR, but optionally, if you want to be extra safe, you could just not update the votes here, and have the user refresh the page to see the updated votes.

anna-giang (approved):
LGTM 👍

GoMeet/src/main/webapp/test/spec/spec-interactive-map.js
```diff
@@ -120,7 +138,7 @@ describe ('Build Info Window Vote', function() {
   const TITLE_A = 'Taco Place';
   const COUNT_A = 2;
   const NOTE_A = 'Tacos taste yum!';
-
+
```
thanhdng: nit: please remove the spaces here

thanhdng (approved):
LGTM w nit

PR #46, opened 2021-02-08
Vote meeting time form display

anna-giang:
JavaScript to dynamically render the form for users to vote for meeting times.

### DIFF
--- a/GoMeet/src/main/webapp/js/meeting-display.js
+++ b/GoMeet/src/main/webapp/js/meeting-display.js
@@ -48,8 +48,8 @@ function getMeetingEventId() {
   let parser = document.createElement('a');
   parser.href = getCurrentURL();
   let query = parser.search.substring(1); // Discard the '?' at position 0 .
-	let vars = query.split('&');
-	for (let i = 0; i < vars.length; i++) {
+  let vars = query.split('&');
+  for (let i = 0; i < vars.length; i++) {
     // Disregard all pairs unless it is the meetingEventId.
     let pair = vars[i].split('=');
     if (pair[0] === 'meetingEventId') { 
--- a/GoMeet/src/main/webapp/js/meeting-display.js
+++ b/GoMeet/src/main/webapp/js/meeting-display.js
@@ -33,3 +33,41 @@ function generateErrorMessage(errorResponse) {
   let message = "ERROR " + errorResponse.status + " " + errorResponse.message;
   return message;
 }
+
+/**
+ * Returns the meetingEventId value from the URL of the current
+ * displayed page. 
+ * Assumes that the URL has a query string of the format: 
+ * meetingEventId=ID
+ * @returns The String meetingEventId retrieved from the URL
+ */
+function getMeetingEventId() {
+  // Create an anchor element with the current URL.
+  // The query string of the URL (INCLUDING the '?') can 
+  // then be accessed via element.search.
+  let parser = document.createElement('a');
+  parser.href = getCurrentURL();
+  let query = parser.search.substring(1); // Discard the '?' at position 0 .
+	let vars = query.split('&');
+	for (let i = 0; i < vars.length; i++) {
+    // Disregard all pairs unless it is the meetingEventId.
+    let pair = vars[i].split('=');
+    if (pair[0] === 'meetingEventId') { 
+      return decodeURIComponent(pair[1]); 
+    }
+  }
+  return null;
+}
+
+/**
+ * Returns the current URL, as stored in window.location.href.
+ * 
+ * This function will make it easier to test functions 
+ * (e.g. getMeetingEventId) that depend on window.location.href - 
+ * mock this function to control the value in window.location.href.
+ * 
+ * @returns The value stored in window.location.href
+ */
+function getCurrentURL() {
+  return window.location.href;
+}
--- a/GoMeet/src/main/webapp/js/meeting-display.js
+++ b/GoMeet/src/main/webapp/js/meeting-display.js
@@ -1,3 +1,6 @@
+/**
+ * Toggles meeting event display depending on login status of user.
+ */
 function toggleMeetingDisplay() {
   LoginStatus.doGet().then((loginStatus) => {
     let prompt; 
@@ -16,4 +19,17 @@ function toggleMeetingDisplay() {
     document.getElementById('votes-table').style.display = style;
     document.getElementById('login-or-logout-prompt').innerHTML = prompt;
   });
+}
+
+/**
+ * Generates an error message string of the format:
+ * "ERROR: errorResponse.status errorResponse.message".
+ * @param {Object} errorResponse The error response object returned from 
+ * a DAO.
+ * @returns The string of the format "ERROR: errorResponse.status 
+ * errorResponse.message"
+ */
+function generateErrorMessage(errorResponse) {
+  let message = "ERROR " + errorResponse.status + " " + errorResponse.message;
+  return message;
 }
--- a/GoMeet/src/main/webapp/js/vote-meeting-time.js
+++ b/GoMeet/src/main/webapp/js/vote-meeting-time.js
@@ -72,4 +72,75 @@ async function fetchAndProcess(meetingTimeIds) {
  * {meetingTimeId: id, datetime: datetimeStr, voteCount: count, voters: [voters]}
  */
 function generateVoteTimeForm(timeData, currentUser, votedTimes) {
+  let votingRight = true;
+  if (votedTimes.size === MAX_VOTES) {
+    votingRight = false; // user may not vote again if they have voted maximum times
+  }
+  // disable all vote buttons for users that have voted MAX_VOTES already
+  let disabledState = !votingRight; 
+
+  sortTimesbyVotes(timeData);
+
+  let voteTable = document.getElementById('meeting-times-table');
+  
+  for (let i = 0; i < timeData.length; i++) {
+    let time = timeData[i];
+    let timeRow = voteTable.insertRow(i + 1);
+
+    let dateTimeCell = timeRow.insertCell(0);
+    dateTimeCell.innerText = time.datetime;
+
+    let voteCountCell = timeRow.insertCell(1);
+    voteCountCell.innerText = time.voteCount;
+
+    let votersCell = timeRow.insertCell(2);
+    votersCell.innerText = time.voters;
+
+    // Clicking on this button will cast a vote for this meeting time
+    let voteFormCell = timeRow.insertCell(3);
+    let voteButton = document.createElement('button');
+    voteButton.id = time.id;
+    voteButton.onclick = function() {voteTime(time.id, currentUser)};
+    if (votedTimes.has(time.id)) {
+      // always disable buttons of times the user has voted for already
+      voteButton.disabled = true;
+    } else {
+      voteButton.disabled = disabledState;
+    }
+    voteFormCell.appendChild(voteButton);
+  }
+}
+
+/**
+ * Sorts the timeData objects given by their 'voteCount'
+ * property in descending order
+ * @param {Array[Object]} timeData an array of timeData objects
+ * retrieved from the '/meeting-time' servlet via GET request, in 
+ * the format:
+ * {meetingTimeId: id, datetime: datetimeStr, voteCount: count, voters: [voters]}
+ */
+function sortTimesbyVotes(timeData) {
+  // sort the times by votes, in decreasing order
+  timeData.sort((time1, time2) => {
+    if (time1.voteCount > time2.voteCount) {
+      return -1;
+    }
+    if (time1.voteCount < time2.voteCount) {
+      return 1;
+    }
+    // voteCounts equal
+    return 0;
+  });
+}
+
+/**
+ * Increments the vote count for the MeetingTime with the id
+ * given. Stores the new vote count, and the email of the user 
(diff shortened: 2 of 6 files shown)


anna-giang:
What the dynamically generated voting form looks like: 

Screen Shot 2021-02-02 at 11 39 13 am

GoMeet/src/main/webapp/js/meeting-display.js
```diff
@@ -17,3 +20,54 @@ function toggleMeetingDisplay() {
...
+  parser.href = getCurrentURL();
+  let query = parser.search.substring(1); // Discard the '?' at position 0 .
+	let vars = query.split('&');
+	for (let i = 0; i < vars.length; i++) {
+    // Disregard all pairs unless it is the meetingEventId.
+    let pair = vars[i].split('=');
+    if (pair[0] === 'meetingEventId') { 
+      return decodeURIComponent(pair[1]); 
+    }
+  }
```
vivianho18: nit - fix indentations

GoMeet/src/main/webapp/js/meeting-display.js
```diff
@@ -17,3 +20,54 @@ function toggleMeetingDisplay() {
...
+  return null;
+}
+
+/**
+ * Returns the current URL, as stored in window.location.href.
+ * 
+ * This function will make it easier to test functions 
+ * (e.g. getMeetingEventId) that depend on window.location.href - 
+ * mock this function to control the value in window.location.href.
```
vivianho18, line 67: Good thinking!

GoMeet/src/main/webapp/meeting-event.html
```diff
@@ -9,6 +9,9 @@
     <script src="js/meeting-display.js"></script>
     <script src="js/login-status.js"></script>
     <link rel="stylesheet" href="style.css">
+    <script src="js/meeting-time-dao.js"></script>
+    <script src="js/vote-meeting-time-dao.js"></script>
+    <script src="js/vote-meeting-time.js"></script>
```
vivianho18, line 13: nit - switch line 13 and 14 so that they are alphabetised

GoMeet/src/main/webapp/meeting-event.html
```diff
@@ -17,7 +20,18 @@ <h2 id="meeting-title">Meeting Name:</h2>
     <div id="vote-meeting-times">
       <h3>Meeting times here</h3>
     </div>
-    <div id="vote-meeting-locations">
+    <h2 id="meeting-name"></h2>
```
vivianho18: This can be deleted - the meeting name will be displayed from line 19

GoMeet/src/main/webapp/test/spec/spec-vote-meeting-time.js
```diff
@@ -0,0 +1,316 @@
+const CURRENT_USER = 'anna@test.com';
+const MEETING_TIME_IDS = ['abc123', 'def456', 'ghi789'];
+// The times that CURRENT_USER has voted for.
+const VOTED_TIMES = new Set(['abc123', 'def456', 'ghi789']);
```
vivianho18: can you pass in MEETING_TIME_IDS here?

GoMeet/src/main/webapp/test/spec/spec-vote-meeting-time.js
```diff
@@ -0,0 +1,316 @@
...
+    spyOn(LoginStatus, 'doGet').and.returnValues(LOGGED_IN, LOGGED_OUT);
+  });
+
+  it('returns the email of the logged in user, \
+      or null if the user is not logged in', async function() {
+    let loggedInUser = await getLoggedInUser();
+    expect(loggedInUser).toEqual(LOGGED_IN.userEmail);
+    let loggedOutUser = await getLoggedInUser();
+    expect(loggedOutUser).toBe(null);
+  })
```
vivianho18: nit - semicolon

GoMeet/src/main/webapp/test/spec/spec-vote-meeting-time.js
```diff
@@ -0,0 +1,316 @@
...
+    generateVoteTimeForm(MEETING_TIME_DATA, "newuser@test.com", votedTimesNewUser);
+    verifyRowData();
+    let table = document.getElementById('meeting-times-table');
+    for (let i = 1; i < table.rows.length; i++) {
+      let dataRow = table.rows.item(i);
+      let voteButton = dataRow.children.item(3).children.item(0);
+      expect(voteButton.disabled).toBe(false); // 'currentUser' has not voted at all
+    }
+  });
+  afterEach(function() {
```
vivianho18, line 194: nit - newline

GoMeet/src/main/webapp/test/spec/spec-vote-meeting-time.js
```diff
@@ -0,0 +1,316 @@
...
+        voters: ['anna@test.com', 'therese@test.com'],
+      },
+      {
+        datetime: '2021-01-27T11:20',
+        voteCount: 3,
+        voters: ['anna@test.com', 'therese@test.com', 'vivian@test.com'],
+      }
+    ];
+    sortTimesbyVotes(timeData);
+    expect(timeData).toEqual(sortedTimeData);
```
vivianho18, line 248: optional - could be good to check a bit more thoroughly. Similar to when tests were passing for the servlet tests even though the convertToJson was returning null. Maybe just a check of the size of timeData

vivianho18 (commented):
lgtm :))

GoMeet/src/main/webapp/test/spec/spec-vote-meeting-time.js
```diff
@@ -0,0 +1,316 @@
...
+        voters: ['anna@test.com', 'therese@test.com'],
+      },
+      {
+        datetime: '2021-01-28T17:20',
+        voteCount: 1,
+        voters: ['anna@test.com'],
+      }
+    ];
+    sortTimesbyVotes(sortedTimeDataDuplicate);
+    expect(sortedTimeDataDuplicate).toEqual(sortedTimeData);
```
vivianho18, line 272: same as above

GoMeet/src/main/webapp/js/vote-meeting-time.js
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * Sorts the timeData objects given by their 'voteCount'
+ * property in descending order
+ * @param {Array[Object]} timeData an array of timeData objects
+ * retrieved from the '/meeting-time' servlet via GET request, in 
+ * the format:
+ * {meetingTimeId: id, datetime: datetimeStr, voteCount: count, voters: [voters]}
+ */
+function sortTimesbyVotes(timeData) {
```
theresenieva, line 146: Perhaps it would be good to explain why you decided to sort the times client-side. :)

GoMeet/src/main/webapp/js/vote-meeting-time.js
```diff
@@ -0,0 +1,174 @@
...
+
+  let timeData = [];
+  let votedTimes = new Set(); // Times that the user has voted for, so cannot vote them again.
+  for (let i = 0; i < meetingTimeIds.length; i++) {
+    let time = await MeetingTimeDAO.fetchMeetingTime(meetingTimeIds[i]);
+    if ('status' in time && parseInt(time.status) !== 200) {
+      // NOTE: Error responses from the DAO are internal errors that the user cannot deal with, 
+      // so don't alert them, but also don't crash the program or stop execution either.
+      // Hence not throwing an error (which stops execution), but logging it.
+      console.error(generateErrorMessage(time) + " - MeetingTimeId: " + meetingTimeIds[i]);
```
theresenieva, line 70: Optional: I feel like if there is a problem with the server then it would be better to alert users. For example, similar to messages when you visit websites that say something like, "Servers down, please try again later."

theresenieva (approved):
lgtm

GoMeet/src/main/webapp/test/spec/spec-vote-meeting-time.js
anna-giang: Yep, I've added a check for the length to be extra careful as you mentioned. 
But should be good in this case since I've already hardcoded the sortedTimeData values in the beforeEach() function, so it's certain that sortedTimeData is not null 👍

GoMeet/src/main/webapp/js/vote-meeting-time.js
anna-giang: Yep, added some comments to do that in the function header. But here's the explanation for easy reference:

Sorting the timeData objects client-side, because it is likely that not every single client-side use case of the timeData requires the data to be sorted in vote order. Hence, sorting the timeData server-side will add unnecessary additional complexity to the server code (and consequently the fetch() calls to the servlet).

GoMeet/src/main/webapp/js/vote-meeting-time.js
anna-giang: The main reason I use `console.error` here instead of throwing an error and alerting the users is that I don't want to stop execution (I'm handling the case where there could be some meetingTimeIds in the array that are invalid, and others that are valid. I want to display all the valid ones, so I don't want to stop execution on any one invalid ID).

I also considered not throwing an error, but simply put an error message as part of the UI (like you mention) for times that couldn't be fetched. But I was thinking for UI neatness sake, it may be better not have the message there, because it comes back to the fact that, even if the user knew that one of the meetingTimes could not be fetched, they wouldn't be able to fix the problem. 
(poor user 😢)

theresenieva: Yup, looks like you've thought about it well. Thanks for the explanation (poor user indeed 😢).

GoMeet/src/main/webapp/js/vote-meeting-time.js
```diff
@@ -0,0 +1,178 @@
...
+async function displayMeetingTimeForm() {
+  let meetingEventId = getMeetingEventId();
+  let meetingEvent = await MeetingEventDAO.fetchMeetingEvent(meetingEventId);
+  let meetingTimeIds = meetingEvent['meetingTimeIds']; 
+
+  if (('status' in meetingEvent && parseInt(meetingEvent.status) !== 200) 
+      || meetingTimeIds === null) {
+    // Execution cannot continue unless meetingTimeIds can be retrieved. 
+    // Must terminate script with an Error. 
+    throw new Error(generateErrorMessage(meetingEvent) + " - MeetingEventId: " + meetingEventId);
```
suprsrsly, line 17: This function throws exceptions and it's called from voteTime() but not wrapped in a try/catch block. What happens if this exception is thrown?

GoMeet/src/main/webapp/js/vote-meeting-time.js
```diff
@@ -0,0 +1,178 @@
...
+ * to the server code (and consequently the fetch() calls to the servlet).
+ * @param {Array[Object]} timeData an array of timeData objects
+ * retrieved from the '/meeting-time' servlet via GET request, in 
+ * the format:
+ * {meetingTimeId: id, datetime: datetimeStr, voteCount: count, voters: [voters]}
+ */
+function sortTimesbyVotes(timeData) {
+  // Sort the times by votes, in decreasing order.
+  timeData.sort((time1, time2) => {
+    if (time1.voteCount > time2.voteCount) {
```
suprsrsly: Seems like you could have done something like `return time2.voteCount - time1.voteCount` instead, is there a reason you did it this way?

GoMeet/src/main/webapp/js/vote-meeting-time.js
```diff
@@ -0,0 +1,178 @@
...
+ * given. Stores the new vote count, and the email of the user 
+ * that voted to Datastore.
+ * @param {String} id The id of the meeting time voted for.
+ * @param {String} currentUser The user that is currently logged in,
+ * and who is voting.
+ */
+async function voteTime(id, currentUser) {
+  let response = await VoteMeetingTimeDAO.voteMeetingTime(id, currentUser);
+  if ('status' in response && parseInt(response.status) !== 200) {
+    // NOTE: Error responses from the DAO are internal errors that the user cannot deal with,
```
suprsrsly, line 164: Just a comment: on the other hand, you might want to notify them that they're getting "degraded service". Hopefully some SRE has set up an alert for when there is a spike in non-200 HTTP responses anyway.

GoMeet/src/main/webapp/js/vote-meeting-time.js
anna-giang: There would be an error in the console _and_ the script execution will terminate, and from the user's perspective, the voting form section on the HTML page would be blank. 

Catching the error will allow execution to continue (but I want termination in this case, and it looks like throwing an error is the [typical way](https://stackoverflow.com/questions/9298839/is-it-possible-to-stop-javascript-execution) to terminate scripts). And once again, the poor user can't fix this issue, so I decided to not notify them via the UI or via `alert()` like I've done in the past.

suprsrsly (approved):

GoMeet/src/main/webapp/js/vote-meeting-time.js
anna-giang: No, I just thought that the number returned had to be -1, 0, or 1 😅 But a closer look at the docs says it's just either negative, 0 or positive, so I've refactored it.

PR #44, opened 2021-02-08
Display popular locations

theresenieva:
Display popular locations on the page.

I've created a servlet to return a JSON array of the popular locations and JavaScript code to display those locations. I chose to display the popular locations as a list below the map when the user clicks a button. I think this would be cleaner than having the list update whenever a new location is added.

PR #47, opened 2021-02-09
Update LocationServlet to check if New Location is Valid

theresenieva:
Update LocationServlet to check if New Location is Valid.

Update locations servlet to send an error response when:
1) The maximum number of locations for a meeting has been reached (5). 
2) There is already a location with the same as title as the new title in that meeting. 

I chose to use custom exceptions in the Dao to signal to that the above has occurred. The servlet will catch these exceptions and send an error response to the client as appropriate. 

The main files for this PR are LocationDao.java and LocationServlet.java. 

Note: 
- I plan to refactor the MeetingLocationDao.js in a seperate PR. I will handle the error responses there. 
- Dependent on PR #44. This PR contains commits from that PR. Commits ```d20495aee3d5676065740fa0c9abd64a0d8f1546``` onwards are from this PR.


GoMeet/src/main/java/com/google/sps/dao/LocationDao.java
```diff
@@ -56,7 +62,16 @@ public LocationDao() {
    * Returns the key string of the new entity.
```
anna-giang: Optional: Javadoc comments should have the @param @return [block tags](https://google.github.io/styleguide/javaguide.html#s7.1.3-javadoc-block-tags) where appropriate.

GoMeet/src/test/java/com/google/sps/LocationServletTest.java
```diff
@@ -88,17 +89,27 @@ public void doPostTest() throws IOException {
...
     // Check the mockedDao was called with the correct parameters
     ArgumentCaptor<Location> captor = ArgumentCaptor.forClass(Location.class);
-    verify(mockedLocationDao, times(1)).save(captor.capture());
-    Location actual = captor.getValue();
-    assertEquals(LOCATION_A, actual);
+
+    try {
+       verify(mockedLocationDao, times(1)).save(captor.capture());
+       Location actual = captor.getValue();
+      assertEquals(LOCATION_A, actual);
```
anna-giang, line 115: nit: fix indent.

GoMeet/src/main/java/com/google/sps/dao/LocationDao.java
```diff
@@ -59,11 +62,15 @@ public LocationDao() {
...
    */
   @Override
-  public String save(Location location) throws SimilarEntityExistsException {
+  public String save(Location location) throws MaxEntitiesReachedException, SimilarEntityExistsException {
```
anna-giang: Optional: It might be good to explain why the case where the maximum number of locations are reached is handled exclusively in the back-end? 
Because there could potentially also be counters in the front-end to see how many locations a user proposes.

GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
```diff
@@ -60,7 +60,8 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       ServletUtil.sendErrorResponse(
           response, HttpServletResponse.SC_BAD_REQUEST, ErrorMessages.BAD_REQUEST_ERROR_LOCATION);
     } catch (MaxEntitiesReachedException e) {
-      // TODO: Handle max entities reached.
+      ServletUtil.sendErrorResponse(
+          response, HttpServletResponse.SC_BAD_REQUEST, ErrorMessages.MAX_ENTITIES);
```
theresenieva, line 64: I would also like some feedback on whether you think this is the right way to signal that the max number of entities has been reached, particularly, whether I've used an appropriate error code.

GoMeet/src/main/java/com/google/sps/dao/LocationDao.java
```diff
@@ -68,7 +90,11 @@ public String save(Location location) {
     return KeyFactory.keyToString(entity.getKey());
   }
 
-  /** Increments the vote count of the location entity with the give keyString. */
+  /** 
+   * Increments the vote count of the location entity with the give keyString.
+   *
```
vivianho18, line 95: nit - remove extra line

GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
vivianho18: Seems appropriate to me

vivianho18 (commented):
lgtm!

GoMeet/src/main/java/com/google/sps/dao/LocationDao.java
```diff
@@ -82,4 +108,35 @@ public void updateVote(String keyString) throws EntityNotFoundException {
...
+   * TODO: Update to use MeetingID.
+   */
+  private boolean validTitle(String targetTitle) {
+    Query q = new Query("Location").setFilter(new FilterPredicate("title", FilterOperator.EQUAL,
+        targetTitle));
+    int numOfSameEntities = ds.prepare(q).countEntities();
+    if (numOfSameEntities > 0) {
+      return false;
+    }
+    return true;
```
ghost: How about `return numOfSameEntities == 0;`?

GoMeet/src/main/java/com/google/sps/dao/LocationDao.java
```diff
@@ -82,4 +108,35 @@ public void updateVote(String keyString) throws EntityNotFoundException {
...
+   * the database for that meeting.
+   * TODO: Update to use MeetingID.
+   */
+  private boolean maxLocationsReached() {
+    Query q = new Query("Location");
+    int numOfEntities = ds.prepare(q).countEntities();
+    if (numOfEntities < MAX_LOCATIONS) {
+      return false;
+    }
+    return true;
```
ghost: Similarly, how about `return numOfEntities >= MAX_LOCATIONS`?

GoMeet/src/main/java/com/google/sps/servlets/LocationServlet.java
ghost: Yeah I don't see anything better

GoMeet/src/test/java/com/google/sps/LocationServletTest.java
```diff
@@ -88,24 +97,98 @@ public void doPostTest() throws IOException {
    
     // Set up DAO mock
     String keyString = KeyFactory.keyToString(KeyFactory.createKey("Location", 123));
-    when(mockedLocationDao.save((Location)notNull())).thenReturn(keyString);
 
-    LocationServlet servlet = new LocationServlet();
-    servlet.setDao(mockedLocationDao);
+    try {
+      when(mockedLocationDao.save((Location)notNull())).thenReturn(keyString);
+    } catch (Exception e) {
```
ghost, line 103: What's the reason for this try catch block? Are you sure this throws an exception? It would seem weird to as it's just mocking out a function. Same with the other new try catches in this file.

ghost (approved):
The pull request's title is a little confusing as it doesn't really tell me what the changes do. It should be an 'action' statement. The commit messages are great though, so you should just treat it like the 'commit message' for the whole PR.

Looks fine except for some comments.

GoMeet/src/main/java/com/google/sps/dao/LocationDao.java
theresenieva: Yup, the explanation is that:

One meeting can have a maximum of 5 locations. If two people are trying to add a location at the same time, there could be the case that:
- Person A and Person B open the event page. If they had front-end counters, it would be at 4. 
- Person A sends in a new location and it gets added to the meeting. Their counter is now 5 but person B's counter is still 4. 
- Person B sends in a new location since their counter still says 4. If there was no server side check, then the meeting would have 6 locations in total. 

I could not implement real-time updating as that was a stretch goal.

GoMeet/src/main/java/com/google/sps/dao/LocationDao.java
theresenieva: Hahah, bit surprised I didn't notice that. Thanks! Will update.

GoMeet/src/test/java/com/google/sps/LocationServletTest.java
theresenieva: Yep, when I ran it without the try-catch, I got the error: Exception needs to be caught or thrown, which is why I added it in. I noticed that this only happened when the method I was mocking threw an exception.

PR #48, opened 2021-02-10
Submit meeting

vivianho18:
Integration of data from the 3-step create meeting sequence. Create relevant meeting event, meeting time and meeting location entities for the event and save all to datastore. Include the unique meeting event entity key in the meeting-event.html page so that fetches can be made to datastore to display the associated data for that meeting. 

Refactoring: 
- Update the login feature to redirect to the meeting-event.html page with the meeting event entity key query string appended i.e. `?meetingEventId=XXX` (NOTE: The case of invalid meeting event ID will be handled in the display meeting functionality so that the user is still able to log in. If the user is redirected to a meeting event page with an invalid ID in the query string, the html page will display an error message). 
- Update the display meeting feature to accept the meeting event entity key as a param
- Call Therese's temp function to get the meeting location ids in newMeetingEvent (NOTE: This function call needs to be updated once Therese implements temp storage for meeting locations)


GoMeet/src/main/webapp/js/login-status.js
```diff
@@ -2,8 +2,17 @@ class LoginStatus {
...
+    if (meetingEventId === null || meetingEventId === undefined) {
+      throw new Error(INSUFFICIENT_REQUEST_PARAM); 
+    }
+
+    if (typeof meetingEventId !== 'string') {
+      throw new Error(INVALID_PARAM_TYPE); 
+    }
+
+    let urlString = DAOUtils.url(LoginStatus.endpoint, {'meetingEventId': meetingEventId});
+    let results = await fetch(urlString).then((results) => results.json());
```
theresenieva, line 15: Have you handled the case where this call to fetch() fails?

theresenieva (approved):
lgtm with one suggestion.

GoMeet/src/main/webapp/js/login-status.js
vivianho18: Added a check in the servlet for bad fetch requests

GoMeet/src/main/webapp/test/spec/spec-login-status.js
```diff
@@ -0,0 +1,47 @@
...
+  }); 
+
+  it('Returns a JSON object with the login status data on success', async function () {
+    let result = await LoginStatus.doGet(MEETING_EVENT_ID);
+    expect(result).toEqual(LOGIN_STATUS_DATA);
+    expect(window.fetch).toHaveBeenCalledWith(
+      LoginStatus.endpoint + QUERY_STRING
+    );
+  });
+});
```
anna-giang: nit: newline

GoMeet/src/main/webapp/js/meeting-display.js
```diff
@@ -1,13 +1,28 @@
...
+        let meetingName = meetingEventEntity['meetingName'];
+        let durationMins = meetingEventEntity['durationMins'];
+        let durationHours = meetingEventEntity['durationHours']; 
+        let timeFindMethod = meetingEventEntity['timeFindMethod']; 
+        let guestList = meetingEventEntity['guestList']; 
+        let meetingTimeIds = meetingEventEntity['meetingTimeIds']; 
+        let meetingLocationIds = meetingEventEntity['meetingLocationIds'];
+        // TO DO: Call neccessary functions to display meeting event data
+        displayMeetingTimeForm(); 
+      }
```
vivianho18: The rationale behind this implementation design is that the user should be able to log in even if they are on a meeting event page with a query string entity key that does not exist. After logging in, the application will verify that a meeting event entity with the given ID exists in datastore and if so, it will call the necessary functions to display the data.

GoMeet/src/main/webapp/js/meeting-display.js
```diff
@@ -1,13 +1,28 @@
 /**
  * Toggles meeting event display depending on login status of user.
  */
-function toggleMeetingDisplay() {
-  LoginStatus.doGet().then((loginStatus) => {
+async function toggleMeetingDisplay() {
+  let meetingEventId = getMeetingEventId();
+  await LoginStatus.doGet(meetingEventId).then((loginStatus) => {
```
ghost: The advantage of `async`/`await` is that you don't need to do `then` chaining here - you could rewrite this as:
```
let loginStatus = await LoginStatus.doGet(meetingEventId);
let prompt;
...
```
This way it reads sequentially and you don't need a new indented block per async call.

ghost (approved):
Looks fine to me :)

GoMeet/src/main/webapp/js/meeting-display.js
```diff
@@ -1,13 +1,28 @@
...
+      let meetingEventEntity = MeetingEventDAO.fetchMeetingEvent(meetingEventId); 
+      if (meetingEventEntity['status'] === undefined) { // No error response was sent 
+        style = 'block';
+        let meetingName = meetingEventEntity['meetingName'];
+        let durationMins = meetingEventEntity['durationMins'];
+        let durationHours = meetingEventEntity['durationHours']; 
+        let timeFindMethod = meetingEventEntity['timeFindMethod']; 
+        let guestList = meetingEventEntity['guestList']; 
+        let meetingTimeIds = meetingEventEntity['meetingTimeIds']; 
+        let meetingLocationIds = meetingEventEntity['meetingLocationIds'];
```
ghost: The above seven variables are unused - is that because of the TO DO next line?

vivianho18: Yes - we would also need to refactor Anna's displayMeetingTimeForm to accept params such as the meetingTimeIds. This is to avoid each of display meeting helper function from having to fetch the meeting entity from datastore in order to access the data. Instead, with this implementation, the fetch is only called once and the data is passed to the functions.

PR #50, opened 2021-02-11
AutoProposeTimes class - Google Calendar Meeting Time Proposal 

anna-giang:
Implementation for the class to that calls the Google Calendar API and generates meeting times.

proposeTimes function uses the following algorithm to analyse a list of TimePeriods representing busy times:

1. Sort the list of TimePeriods in order of earliest to latest start time
2. Scanning over the list of TimePeriod with two  pointers:
    If the end of the current time period (pointed to by left pointer) overlaps with the start of the next time period, move the right pointer one item to the right.
    Otherwise, check the duration of the free time block from the end of the current time period to the start of the next time period.
        If the duration is at least the meeting duration, add to output.
        Move the left pointer to the right pointer (i.e. current is now what the right pointer was pointing to.)

The algorithm time complexity is bounded by the sort of the time periods, O(Nlog(N)), where N is the number of busy TimePeriods in total retrieved from each of the invitees' calendars.

### DIFF
--- a/GoMeet/src/main/java/com/google/sps/data/AutoProposeTimes.java
+++ b/GoMeet/src/main/java/com/google/sps/data/AutoProposeTimes.java
@@ -1,6 +1,7 @@
 package com.google.sps.data;
 
 import com.google.api.client.util.DateTime;
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.services.calendar.Calendar.Freebusy.Query;
 import com.google.api.services.calendar.Calendar.Freebusy;
 import com.google.api.services.calendar.Calendar;
@@ -12,11 +13,19 @@ import com.google.api.services.calendar.model.FreeBusyResponse;
 import com.google.api.services.calendar.model.TimePeriod;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
+import java.util.Map;
 
 public class AutoProposeTimes {
+  private final String RFC3339_FORMAT = "yyyy-MM-dd'T'HH:mm:ssXXX";
+  private final String API_KEY = ""; // API KEY HERE
+
   Calendar service;
   ArrayList<String> calendarId; // The List of Calendar IDs to generate meeting times for
   DateTime startTime; // The start time of the period to look for possible meeting times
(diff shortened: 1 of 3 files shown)


GoMeet/src/main/java/com/google/sps/data/AutoProposeTimes.java
```diff
@@ -0,0 +1,181 @@
+package com.google.sps.data;
+
+import com.google.api.client.util.DateTime;
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
```
vivianho18, line 3: nit - alphabetise imports

GoMeet/src/main/java/com/google/sps/data/AutoProposeTimes.java
```diff
@@ -0,0 +1,180 @@
...
+    durationEndOfDay = this.endTime.getValue() - lastBusyPeriodEnd.getValue();
+    if (durationEndOfDay >= this.meetingDuration) {
+      freePeriods.add(newTimePeriod(lastBusyPeriodEnd, this.endTime));
+    }
+
+    return freePeriods;
+  }
+
+  /**
+   * Makes a request to the Freebusy library of the Google Calendar API,
```
theresenieva, line 143: With making requests to an external API, is there a chance that the request could not be processed? And should that be handled?

This may not be so important because I understand this is more to demonstrate the algorithm for finding free periods so just something to think about.

theresenieva (commented):
lgtm :) I like the explanation with the time complexity.

GoMeet/src/main/java/com/google/sps/data/AutoProposeTimes.java
anna-giang: Yep, I've handled the case (see [PR 51](https://github.com/googleinterns/step271-2020/pull/51)) where there are security error responses from the API. There will be an error thrown (GeneralSecurityError or GoogleJsonResponseException) by these functions that call the API, and I'm planning to have the servlet that calls these functions handle that by returning an error response to the user like we usually do.

I'm sure there are other errors (invalid request format, etc.) that I haven't explicitly handled, so I think I might modify my servlet class to just catch all exceptions and return the error message in an error response to the front-end. Thanks for your suggestion

GoMeet/src/main/java/com/google/sps/data/AutoProposeTimes.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    // Generate the free periods: O(N)
+    int i = 0;
+    TimePeriod current = busyPeriods.get(i);
+    DateTime lastBusyPeriodEnd = current.getEnd();
+
+    while (i < busyPeriods.size()) {
+      int j = i + 1;
+      while (j < busyPeriods.size()
```
suprsrsly, line 109: Another way you could do this is by sorting all the start and end times independently, traverse the list from start to end, and keep a counter for how many events you're "currently in". When the counter is 0, it means you've got some free time. e.g.,

0800 - start of day; numCurrentEvents = 0
0900 - start meeting 1; numCurrentEvents = 1 --> append(new freeTime(0800, 0900))
1000 - start meeting 2; numCurrentEvents = 2
1100 - end meeting 1; numCurrentEvents = 1
1200 - end meeting 2; numCurrentEvents = 0
1700 - end of day --> append(new freeTime(1200, 1700))

If you treat start of day as the "end" of an event, and the end of the day as the "start" of an event, your beginning-of-day and end-of-day edge cases are also eliminated.

suprsrsly: note this is optional, you already have something working.

suprsrsly (approved):

GoMeet/src/main/java/com/google/sps/data/AutoProposeTimes.java
anna-giang: This is a smart alternative - I didn't think to include the start and end times like that 😅 For the times sake and in case I break anything, I'm going to leave what I have, but thank you for the feedback!

PR #51, opened 2021-02-11
GoogleCalendarTimesServlet and tests

anna-giang:
This will be the servlet to access the Google Calendar auto-generated meeting times.

### DIFF
--- a/GoMeet/src/main/java/com/google/sps/servlets/GoogleCalendarTimesServlet.java
+++ b/GoMeet/src/main/java/com/google/sps/servlets/GoogleCalendarTimesServlet.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
+import java.util.TimeZone;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
@@ -27,6 +28,9 @@ import javax.servlet.http.HttpServletResponse;
 
 @WebServlet("/gcal-find-times")
 public class GoogleCalendarTimesServlet extends HttpServlet {
+  // The timezone that the servlet will treat PERIOD_START and PERIOD_END in
+  private static String TIMEZONE = "AET"; // Australian Eastern Time by default
+
   // TODO: Load the API_KEY from a file.
   private static String API_KEY = "";
   
@@ -74,6 +78,7 @@ public class GoogleCalendarTimesServlet extends HttpServlet {
     
     try {
       SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm");
+      formatter.setTimeZone(TimeZone.getTimeZone(TIMEZONE));
       periodStart = formatter.parse(periodStartStr);
       periodEnd = formatter.parse(periodEndStr);
     } catch (ParseException e) {
@@ -108,7 +113,7 @@ public class GoogleCalendarTimesServlet extends HttpServlet {
     // Return times that are the start times of each TimePeriod in times
     List<String> timesStr = new ArrayList<String>();
     for (int i = 0; i < times.size(); i++) {
-      String current = times.get(i).getStart().toString();
+      String current = times.get(i).getStart().toStringRfc3339();
       timesStr.add(current);
     }
     Gson gson = new Gson();
@@ -133,6 +138,9 @@ public class GoogleCalendarTimesServlet extends HttpServlet {
   public List<TimePeriod> autoProposeTimes(HttpServletResponse response, 
       ArrayList<String> guestList, Date periodStart, 
       Date periodEnd, int totalDurationMs) throws IOException {
+    
+    // Ensure that AutoProposeTime is working in the same timezone as this servlet.
+    AutoProposeTimes.setTimezone(TIMEZONE); 
     
     AutoProposeTimes proposer;
     List<TimePeriod> times;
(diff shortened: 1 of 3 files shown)


GoMeet/src/main/java/com/google/sps/servlets/GoogleCalendarTimesServlet.java
```diff
@@ -0,0 +1,167 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/gcal-find-times")
+public class GoogleCalendarTimesServlet extends HttpServlet {
+  // TODO: Load the API_KEY from a file.
+  private static String API_KEY = "";
+  
+  /** Field names used for accessing pairs in the query string 
+    * sent to this servlet */
+  public static class QueryStringFieldNames {
```
theresenieva, line 39: optional: Since I believe this class isn't being referred to outside of your tests, you could make this private.

GoMeet/src/test/java/com/google/sps/GoogleCalendarTimesServletTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+    // Check that the response is the expected.
+    List<String> expectedReturn = Arrays.asList(
+        // The start times of the free time blocks.
+        "2021-02-11T10:00:00.000Z",
+        "2021-02-11T14:00:00.000Z"
+    );
+    Gson gson = new Gson();
+    writer.flush();
+    assertTrue(stringWriter.toString().contains(gson.toJson(expectedReturn)));
```
theresenieva: nit: Similar to how when testing if a hash-map is correctly sent, we convert the response to a hash-map and check values, I think it would be good to convert stringWriter.toString() to a list and check if it contains the correct strings.

theresenieva (approved):
lgtm :)

GoMeet/src/main/java/com/google/sps/servlets/GoogleCalendarTimesServlet.java
anna-giang: It's referred to in the test cases 😅

theresenieva: Oh my, oops. I must have missed that. Thanks for correcting me.

anna-giang: No worries! It's all good 😊

GoMeet/src/main/java/com/google/sps/servlets/GoogleCalendarTimesServlet.java
```diff
@@ -0,0 +1,185 @@
...
+    
+    try {
+      SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm");
+      formatter.setTimeZone(TimeZone.getTimeZone(TIMEZONE));
+      periodStart = formatter.parse(periodStartStr);
+      periodEnd = formatter.parse(periodEndStr);
+    } catch (ParseException e) {
+      ServletUtil.sendErrorResponse(response, 
+          HttpServletResponse.SC_BAD_REQUEST, 
+          ErrorMessages.INVALID_ARG_FORMAT);
```
suprsrsly: I presume that periodStartStr and periodEndStr are both generated by the front end and do not come from the user? (and, I suppose the frontend is validating anything the user enters anyway?)

Just my opinion, but If so, from a user perspective, I wonder what benefit it adds to tell them that the argument format is invalid rather than say, the backend got a bad request like on line 63? The backend could log the problem so that a sysadmin/developer could come in and see what's going on, but I suspect that from the user's point of view, they don't really have control over the argument format.

anna-giang: Yep, both the strings are generated by front end code. I was thinking the more specific error message might help identify which part of the query string was 'bad', which might help with debugging the front end code. It wasn't intended as a message for users, more for the front end developer. 

But I guess it would be justifiable to just send a generic error message - developer will still get a reasonable understanding of what went wrong. Or, I could also just throw the ParseException error rather than sending an error response - either way, the developer will get the message that there is a backend error.

GoMeet/src/main/java/com/google/sps/servlets/GoogleCalendarTimesServlet.java
```diff
@@ -0,0 +1,185 @@
...
+    
+    try {
+      Calendar service = new Calendar.Builder(GoogleNetHttpTransport.newTrustedTransport(),
+        GsonFactory.getDefaultInstance(), 
+        null)
+        .setApplicationName("GoMeet")
+        .build();
+      proposer = new AutoProposeTimes(service, API_KEY, guestList, 
+          periodStart, periodEnd, totalDurationMs);
+      times = proposer.proposeTimes();
```
suprsrsly, line 156: Just for conversation, in hindsight, it might not have been necessary to create an object with internal state to do this since the usage model is instantiate, execute, destroy. I wonder if it would have sufficed just used a singe static function to take all the dependencies and arguments instead.

GoMeet/src/main/java/com/google/sps/servlets/GoogleCalendarTimesServlet.java
suprsrsly: btw changing this is optional.

suprsrsly (approved):

GoMeet/src/main/java/com/google/sps/servlets/GoogleCalendarTimesServlet.java
anna-giang: Yep, OK 👍

GoMeet/src/main/java/com/google/sps/servlets/GoogleCalendarTimesServlet.java
anna-giang: Hmm yeah I think that can be something to refactor in the future.

PR #52, opened 2021-02-12
Display Location Data on a Chart

theresenieva:
Display Location Data on a Chart.

I've added buttons to the meeting page to allow users to switch between the map view and the chart view of the location data.

### DIFF
--- a/GoMeet/src/main/webapp/js/interactive-chart.js
+++ b/GoMeet/src/main/webapp/js/interactive-chart.js
@@ -1,6 +1,6 @@
 google.charts.load('current', {'packages':['corechart']});
 
-/** Fetches fruit data and uses it to create a chart. */
+/** Fetches location data and uses it to create a chart. */
 async function drawChart() {
   const dao = MeetingLocationDaoFactory.getLocationDao('permanent');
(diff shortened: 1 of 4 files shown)


theresenieva:
Screen Shot 2021-02-12 at 10 38 10 am

Screen Shot 2021-02-12 at 10 38 20 am

GoMeet/src/main/webapp/js/interactive-chart.js
```diff
@@ -0,0 +1,21 @@
+google.charts.load('current', {'packages':['corechart']});
+
+/** Fetches fruit data and uses it to create a chart. */
```
anna-giang: This comment should be updated 🍎

anna-giang (approved):
Looks good to me 👍 (reviewed the last 3 commits in this PR as you mentioned)

GoMeet/src/main/webapp/js/interactive-chart.js
theresenieva: Good catch! Thanks for pointing it out

ghost (approved):
LGTM

PR #53, opened 2021-02-12
Add HTML page for Google Calendar time proposal selection 

anna-giang:
Added HTML page and JavaScript for the page where users that selected to use Google Calendar to generate meeting times will input the time period to generate times for, and view the automatically proposed times.


anna-giang:
Screenshot of the new HTML page

Screen Shot 2021-02-12 at 3 48 54 pm

GoMeet/src/main/webapp/create-meeting-step2-gcal.html
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
+<html>
+  <!-- Create Meeting Step 2 GCal: Automatically propose times with Google Calendar -->
+  <head>
+    <meta charset="UTF-8">
+    <title>Create Meeting</title>
+    <script src="js/error-messages.js"></script>
+    <script src="js/time-proposal-util.js"></script>
+    <script src="js/dao-utils.js"></script>
```
theresenieva, line 9: nit: Alphabetical order

GoMeet/src/main/webapp/js/gcal-time-proposal.js
```diff
@@ -0,0 +1,152 @@
...
+ * AEDT, and if end is later than start. Otherwise returns false.
+ */
+function rectifyTimeBoundaries(start, end) {
+  // Verify that both start and end are in the future, and
+  // that end is later than start by at least 1 minute.
+  let startDate = new Date(start);
+  let endDate = new Date(end);
+  if (TimeProposalUtil.verifyFutureTime(start) &&
+      TimeProposalUtil.verifyFutureTime(end) &&
+      endDate > startDate) {
```
theresenieva: Could this be simplified to: ```return (TimeProposalUtil.verifyFutureTime(start) && TimeProposalUtil.verifyFutureTime(end) && endDate > startDate);```

GoMeet/src/main/webapp/js/gcal-time-proposal.js
```diff
@@ -0,0 +1,152 @@
...
+ * server.
+ * @returns {List[String]} The list of times displayed to the user,
+ * or null if the serverResponse was an error response.
+ */
+function displayAutoGeneratedTimes(serverResponse) {
+  let displayDiv = document.getElementById('generated-times');
+  // Clear whatever is in the <div>
+  displayDiv.innerHTML = '';
+
+  if ('status' in serverResponse && parseInt(serverResponse.status) !== 200) {
```
theresenieva: Optional: I feel like this could use two helper functions. One to display error message and the other to display the times.

GoMeet/src/main/webapp/js/time-proposal-util.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Verifies that the time is a time in the future from now,
+   * AEDT time.
+   * @param {String} time The datetime string to be verified.
+   * @returns True if the time is in the future, false otherwise.
+   */
+  static verifyFutureTime(time) {
+    let timeEntered = new Date(time);
+    let now = new Date(TimeProposalUtil.getDatetimeNow());
+    if (!(timeEntered > now)) {
```
theresenieva: This could also be simplified :)

theresenieva (approved):
lgtm :))))

GoMeet/src/main/webapp/create-meeting-step2-gcal.html
anna-giang: I've doing a little bit of [research](https://eager.io/blog/everything-I-know-about-the-script-tag/) into the order that <script> tags need to be in, and it looks like script tags will execute in the order they are loaded, so scripts that are dependent on other scripts should be loaded later (rather than in alphabetical order like in Java).

Because most of our JavaScript are in functions that aren't called until a DOM event happens (by the time when all scripts are loaded) the order in which they appear is less important (so I _can_ rearrange in alphabetical order and no errors occur), but I think because all the <scripts> below error-messages.js use constants defined there, error-messages.js should be loaded first in case of errors that get introduced later when (_assuming that 😅_) changes are made to the code.

Similarly, both dao-utils and time-proposal-utils are used in the <script> tags listed below.

GoMeet/src/main/webapp/js/gcal-time-proposal.js
anna-giang: Yep! Thanks for pointing that out 😊

GoMeet/src/main/webapp/create-meeting-step2-gcal.html
theresenieva: Ah okay. Looks like you've thought about it quite a bit so sounds good :)

anna-giang:
Rebasing on to main to sync with new changes 😊

GoMeet/src/main/webapp/js/gcal-time-proposal.js
anna-giang: I've refactored the error message display to a separate function, because that is more readable 😊 But I've left the code to display the actual times in this function, because given the name of this function `displayAutoGeneratedTimes` I think that code is rightfully left there

anna-giang:
Rebasing on to main again, to sync with new changes.

GoMeet/src/main/webapp/js/gcal-time-proposal.js
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Returns true or false whether start and end times are in the future, 
+ * and whether end is later than start.
+ * @param {String} start The start time, in the format yyyy-MM-ddTHH:mm.
+ * @param {String} end The end time, in the format yyyy-MM-ddTHH:mm
+ * @return True if both start and end are times in the future from now, 
+ * AEDT, and if end is later than start. Otherwise returns false.
+ */
+function rectifyTimeBoundaries(start, end) {
```
suprsrsly: nit: The use of the verb 'rectify' is sort of misleading for a function that returns a boolean. Makes it seem like it modifies state.

GoMeet/src/main/webapp/js/gcal-time-proposal.js
suprsrsly: Just a comment: note that successful status codes can be anywhere from 200 to 299.

GoMeet/src/main/webapp/js/time-proposal-util.js
```diff
@@ -0,0 +1,76 @@
...
+   * @param {Set} enteredTimes The set of datetime strings representing
+   * all the times entered so far. Function will add 'time' to this
+   * set if 'time' is unique and in the future.
+   * @return true if the time is unique and in the future,
+   * otherwise false
+   */
+  static verifyUniqueFutureTime(time, enteredTimes) {
+    if (TimeProposalUtil.verifyFutureTime(time) && 
+        !(enteredTimes.has(time))) {
+      enteredTimes.add(time);
```
suprsrsly, line 56: nit: I would call this function "addFutureTime" or similar instead, since the state-modifying effect of it is to add it to enteredTimes. Calling it 'verify...' makes it seems like the function doesn't modify state.

GoMeet/src/main/webapp/js/time-proposal-util.js
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Clears the input field and alerts the user, if the user entered a time that
+   * is earlier than or equal to the date and time now. See verifyUniqueFutureTime.
+   * @param {Element} elem the input element where the user inputted the time
+   * @param {Set} enteredTimes The set of datetime strings representing
+   * all the times entered so far. Function will add 'time' to this
+   * set if 'time' is unique and in the future.
+   */
+  static rectifyInputtedTime(elem, enteredTimes) {
```
suprsrsly, line 36: nit: this function name is misleading because its purpose is to add a time to the enteredTimes array, not modify the time that's been entered (I don't think...)

suprsrsly (approved):
Mostly good, but some stylistic niggles that I suggest you consider. Please address the function naming comments before submitting.

GoMeet/src/main/webapp/js/gcal-time-proposal.js
anna-giang: Yep, I agree. I've changed it.

GoMeet/src/main/webapp/js/time-proposal-util.js
anna-giang: Yep, that makes sense 👍

GoMeet/src/main/webapp/js/time-proposal-util.js
anna-giang: This function will clear what the user inputted in the <input> element in the 'elem' parameter, so I think it does 'rectify' the time.

PR #55, opened 2021-02-15
Handle bad request response status

theresenieva:
Handle Bad Request Response Status. 

An alert is created when a response is received and the Response.status is not OK. 

I chose to check the response.status in the ```PermMeetingLocationDao``` class so that if there are any changes to how the data is sent to the server/data is received from the server, the functions that update the DOM (in  ```interactive-map.js```) do not have to be changed as they only catch an error.

While, for now, I think handling errors the same way is okay, the handleError() function could be updated to display errors differently depending on the type.

### DIFF
--- a/GoMeet/src/main/java/com/google/sps/data/ErrorMessages.java
+++ b/GoMeet/src/main/java/com/google/sps/data/ErrorMessages.java
@@ -12,7 +12,7 @@ public class ErrorMessages {
       "Too many results returned";
   public static final String REPEAT_TITLE_ERROR = 
       "There is already a location with this title";
-  public static final String MAX_ENTITIES = 
+  public static final String MAX_ENTITIES_ERROR = 
       "Maximum number of entities reached";
   public static final String BAD_EMAIL_REQUEST_ERROR = 
       "MeetingId and guest list must both be provided to send email invitations";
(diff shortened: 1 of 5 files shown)


GoMeet/src/main/java/com/google/sps/data/ErrorMessages.java
```diff
@@ -11,7 +11,7 @@
   public static final String TOO_MANY_RESULTS_ERROR = 
       "Too many results returned";
   public static final String BAD_REQUEST_ERROR_LOCATION =
```
anna-giang: Optional: The name of the error message constant may need to be made more specific if the message is specific.

anna-giang: [Side note: The way that error messages are grouped and named is something I've been thinking would've needed refactoring if we had more time - e.g. whether to have a set of generic error messages that are shared between all the code (fewer error message constants, easier to manage, but less useful for the debugger), or very specific ones that can only be used by one or two functions (may be more difficult to update later if message needs to be changed). Because currently we have a combination of both, but they're all in one shared file which is confusing.]

anna-giang (approved):
LGTM! 😊

GoMeet/src/main/java/com/google/sps/data/ErrorMessages.java
theresenieva: Yeah, you're right. Personally, I think we should have made the error messages more generic. Some of these messages could be combined. For example, a few of them are used for incorrect request parameters (which is something that can happen at any of the servlets).

thanhdng (approved):
LGTM

